PR #2, opened 2019-06-03
Add OpenOCD configuration

hannobraun:
With this configuration, an example can be run on the Nucleo board like this:
```
cargo run --example=blinky --features=rt,stm32f746
```


mvertescher:
Thanks again @hannobraun! LGTM!

PR #3, opened 2019-06-04
Fix linker error in `hello` example

hannobraun:
It didn't link to the HAL or PAC, so the vector table wasn't included in
the binary. This lead to an error in the `cortex-m-rt` linker script.


mvertescher:
Yup, good catch! Thanks!

PR #5, opened 2019-06-06
Fix USART baud rate selection

hannobraun:
The system clock wasn't actually selected as the USART clock source,
except for USART1. This extends the macro to select the correct clock
source for all USART instances.


mvertescher:
LGTM, thanks for the fix!

PR #7, opened 2019-06-07
Add oversampling configuration for USART

hannobraun:


mvertescher:
Neat, thanks @hannobraun!

PR #12, opened 2019-06-14
Add DMA support for USART

hannobraun:
I've started implemented DMA support for USART. I'm submitting this as a draft, as it's not done yet.

Currently, only writing to USART via DMA is supported. Next week, I plan to implement reading from USART and add support for DMA interrupts.

@mvertescher If you have any thoughts so far, I'd love to hear them. I realize this is a complex API, but I believe the complexity is required to prevent incorrect use of the DMA peripheral through the HAL API.

### DIFF
--- a/examples/serial_dma.rs
+++ b/examples/serial_dma.rs
@@ -1,0 +1,73 @@
+//! Write a string to the serial port every half second using DMA.
+//!
+//! Note: This example is for the STM32F746
+
+
+#![deny(unsafe_code)]
+#![deny(warnings)]
+
+#![no_main]
+#![no_std]
+
+
+extern crate panic_halt;
+
+
+use cortex_m_rt::entry;
+use stm32f7xx_hal::{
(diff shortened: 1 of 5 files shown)


hannobraun:
I've changed my mind. I'm submitting this pull request as-is. The extensions I'm working on are a bit more involved than I though, and there's really no point in overloading one pull request with all of them, when this is already useful as-is.

mvertescher:
Spent a bit reviewing this PR and overall it seems consistent with `stm32f7` family and other existing DMA implementations. As a result, I don't have any major concerns. Also, I didn't find the API too complex, however we may find ways to simplify it in the future after some use.

> The extensions I'm working on are a bit more involved than I thought, and there's really no point in overloading one pull request with all of them, when this is already useful as-is.

Sounds like a good plan to me. This PR certainly provides value already.

I definitely like where this implementation is heading. Thanks for your work on this @hannobraun!

PR #14, opened 2019-06-18
Add support for reading from USART using DMA

hannobraun:
This largely follow this guide from The Embedonomicon:
https://docs.rust-embedded.org/embedonomicon/dma.html


mvertescher:
Great to see RX DMA implemented, thanks @hannobraun!

PR #23, opened 2019-11-02
Fix code formating for crate

no111u3:
Change project formating to common format style.
I think it need for automatic merge between existing features and this library code.


mvertescher:
Nice clean up, thanks @no111u3 !

PR #26, opened 2019-11-05
Clean warnings for Rcc module of crate

no111u3:


mvertescher:
Thanks for fixing this up!

PR #27, opened 2019-12-06
Fix set_open_drain for alternate mode gpios

rfuest:
`set_open_drain` used the `&` operator instead of the correct `|` operator. The corrected code matches the behavior of `into_open_drain_output`.


mvertescher:
LGTM!

PR #36, opened 2020-03-18
Support more than just the 746, some changes stolen from the 4xx hal

dtjones190:
From stm32f4xx-hal:
GPIO Configuration
Method to clear timer interrupts

Added a separate timer clock to the RCC clocks since they are sometimes different that the peripheral clocks.

Added conditional compilation so that hal will work with more than just the 746.
Tested compiling with all the listed devices.
Only tested on hardware with a stm32f767 because that is all I have.


src/lib.rs
```diff
@@ -3,21 +3,7 @@
...
-    feature = "stm32f746",
-    feature = "stm32f756",
-    feature = "stm32f765",
-    feature = "stm32f767",
-    feature = "stm32f769",
-    feature = "stm32f777",
-    feature = "stm32f778",
-    feature = "stm32f779",
-)))]
+#[cfg(not(feature = "device-selected"))]
```
mvertescher: :+1:

Cargo.toml
```diff
@@ -43,20 +43,22 @@ panic-halt = "0.2.0"
 panic-semihosting = "0.5.2"
 
 [features]
+device-selected = []
```
mvertescher: Awesome, lgtm!

PR #40, opened 2020-04-20
Add basic support for LTDC screen controller

grodino:
Addded:
* LTDC config
* Layers config

Tested on STM32F7DISCO board with embedded graphics with pixel formats L8 and RGB565.

Question : The LTDC periph needs and address and a length to know where to find the FrameBuffer(s). For now a frame buffer is just a ` &'static mut [T]` with `T: SupportedWord` but I'd like to make it more general.
The DMA already includes some buffer trait and struct, should I put it in a separate module or use it from ltdc ?


mvertescher:
> The LTDC periph needs and address and a length to know where to find the FrameBuffer(s). For now a frame buffer is just a &'static mut [T] with T: SupportedWord but I'd like to make it more general.
>The DMA already includes some buffer trait and struct, should I put it in a separate module or use it from ltdc ?

Good question. IMO seems reasonable to share `Buffer`/`PtrBuffer` with `ltdc` and it would probably be clearer if we broke out those types into their own module (`buffer.rs` or something). Also, I'm cool if we do that in this patch or save it for future work. Let me know what you think @grodino ?

src/ltdc.rs
```diff
@@ -0,0 +1,323 @@
...
+            2   => 0b00,
+            4   => 0b01,
+            8   => 0b10,
+            16  => 0b11,
+            _   => unreachable!()
+        };
+        
+        // // Write PPLSAI configuration
+        rcc.pllsaicfgr.write(|w| unsafe{
+            w   .pllsain().bits(best_plln as u16)
```
mvertescher, line 121: Some weird spacing here... fyi you can run `cargo fmt` on this file if you want

grodino: Allright, I'll do that

grodino:
I do have an example but it is quite messy for now. I'm working on hardware acceleration now, I will clean up the example after. 
You can merge now or then, as you want !

src/ltdc.rs
grodino: Allright, I commit it

mvertescher:
> I do have an example but it is quite messy for now. I'm working on hardware acceleration now, I will clean up the example after.

Sounds good!

PR #41, opened 2020-04-24
New example and basic hardware acceleration

grodino:
Hi !
So in this one I added a new example to demonstrate the ltdc hal and some basic hardware acceleration.


mvertescher:
Thanks for following up with an example for this @grodino ! :100:

PR #43, opened 2020-05-02
Add i2c, correct ltdc and complete rcc

grodino:
In this PR :
* Add i2c : basic support of i2c in standart mode (what is left to support Fast and Fast Plus modes is clock configuration)
* correct ltdc : there was a confusion in the hardware acceleration method that is now fixed
* complete rcc : added Reset and Enable traits for peripherals in order to simplify the "turn that peripheral's clock on" stage. The job now is to find which peripheral is supported by which processor.

Those changes where greatly inspired by the stm32f1xx hal

### DIFF
--- a/src/i2c.rs
+++ b/src/i2c.rs
@@ -1,0 +1,550 @@
+//! Inter-Integrated Circuit (I2C) bus
+//! For now, only master mode is implemented
+
+// NB : this implementation started as a modified copy of https://github.com/stm32-rs/stm32f1xx-hal/blob/master/src/i2c.rs
+#![allow(dead_code)]
+#![allow(unused_macros)]
+#![allow(unused_imports)]
+
+use micromath::F32Ext;
+
+use crate::time::Hertz;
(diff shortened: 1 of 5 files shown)


pftbest:
Have you seen this kind of warnings on this commit?
```
warning: floating-point types cannot be used in patterns
   --> lib/stm32f7xx-hal/src/i2c.rs:323:39
    |
323 |                                   0.0..=16.0 => sclh as u8 - 1,
    |                                         ^^^^
...
547 | / hal! {
548 | |     I2C1: (_i2c1),
549 | |     I2C2: (_i2c2),
550 | | }
    | |_- in this macro invocation
    |
    = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
    = note: for more information, see issue #41620 <https://github.com/rust-lang/rust/issues/41620>
    = note: this warning originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)
```

Maybe it would be better to replace it with integer matches?

grodino:
I don't recall that but sure, we should fix it !

PR #45, opened 2020-06-21
Add support for STM32F730

rfuest:
This PR adds support for the STM32F730. I did test the `blinky` example on real hardware and everything seems to work as expected, but I haven't tested any other peripherals yet.

Because the controller wasn't supported by the `stm32f7` crate version 0.9.0 I had to bump the dependency to the current version 0.11.0.


mvertescher:
Just testing `blinky` sounds good for now. We can fix up any issues with the peripherals as they appear.

Thanks for this patch @rfuest !

PR #46, opened 2020-06-21
Use delay_us from stm32f4xx-hal crate

rfuest:
This PR replaces the current `delay_us` implementation with the one from the `stm32f4xx-hal` crate. The old implementation caused a panic when the value for `us` was too large. The updated implementation doesn't panic and starts the timer multiple times if necessary to achieve the desired delay.


mvertescher:
Awesome, thanks for fixing this @rfuest!

PR #48, opened 2020-06-22
Remove hse field from Clocks struct

rfuest:
The `hse` field in the `Clocks` struct caused a compiler warning, because it wasn't used anywhere. Because the HSE frequency isn't directly used by any peripheral I decided to remove the field entirely.

I did run `rustfmt` on the file to make the formatting more consistent, but this removed some special formatting on some `match` blocks. IMO it's better to have consistent formatting than to maintain the manual formatting, but if you prefer I can also prepare a version of the PR that doesn't contain these changes.

### DIFF
--- a/src/rcc.rs
+++ b/src/rcc.rs
@@ -36,9 +36,9 @@ impl RccExt for RCC {
 pub struct Rcc {
     /// Advanced High-Performance Bus 1 (AHB1) registers
     pub ahb1: AHB1,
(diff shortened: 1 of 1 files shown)


mvertescher:
> I did run rustfmt on the file to make the formatting more consistent, but this removed some special formatting on some match blocks. IMO it's better to have consistent formatting than to maintain the manual formatting

Totally agree :smile: .

PR #51, opened 2020-06-23
Enable DMA and SPI for all devices

rfuest:
The PR enables DMA and SPI support for all devices. The `dma-support` feature is no longer necessary, because all STM32F7 have a DMA controller.

I haven't tested DMA, because I was mainly interested in SPI support. But I've checked that the crate compiles for all devices and SPI was confirmed to work on the previously unsupported STM32F730.


mvertescher:
Cool, thanks for enabling this! This will be helpful for folks to test DMA on other devices and surface any issues. :+1:

PR #52, opened 2020-06-25
Add support for flash programming

rfuest:
This PR adds a `Flash` struct which can be used to program the flash memory. Currently only devices that don't have dual bank flash are supported.


examples/flash.rs
```diff
@@ -0,0 +1,46 @@
+//! Erases a flash sector and programs data.
```
mvertescher, line 1: :+1: super helpful!

mvertescher:
Sweet, thanks for this feature!

PR #59, opened 2020-07-04
Add configuration to RCC PLL Q output with 48 MHz Mode

no111u3:
Changed RCC PLL configuration: Adding PLL Q output configuration for support 48 MHz Mode - needed by USB block.

### DIFF
--- a/src/rcc.rs
+++ b/src/rcc.rs
@@ -352,7 +352,7 @@ impl CFGR {
             let pllq = (vco_in * plln + 47_999_999) / 48_000_000;
             pll48clk = Some(Hertz(vco_in * plln / pllq));
 
-            unsafe { &*RCC::ptr() }.pllcfgr.write(|w| unsafe {
+            rcc.pllcfgr.write(|w| unsafe {
                 w.pllm().bits(pllm as u8);
                 w.plln().bits(plln as u16);
                 w.pllp().bits(pllp as u8);
--- a/src/rcc.rs
+++ b/src/rcc.rs
@@ -165,6 +165,24 @@ impl HSEClock {
 
 const HSI: u32 = 16_000_000; // Hz
 
+#[derive(Default)]
+pub struct RCCRawConfig {
+    pub pllm: u8,
+    pub plln: u16,
+    pub pllp: u8,
+    pub pllq: u8,
+    pub hpre: u8,
+    pub ppre1: u8,
+    pub ppre2: u8,
+    pub flash_waitstates: u8,
+    pub hclk: u32,
+    pub pclk1: u32,
+    pub pclk2: u32,
+    pub sysclk: u32,
+    pub timclk1: u32,
+    pub timclk2: u32,
+}
+
 pub struct CFGR {
     hse: Option<HSEClock>,
     hclk: Option<u32>,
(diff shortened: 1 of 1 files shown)


src/rcc.rs
```diff
@@ -286,91 +293,75 @@ impl CFGR {
...
-                });
-            }
+                sysclk * sysclk_div / vco_in
+            };
+            let pllp = (sysclk_div / 2) - 1;
+
+            let pllq = (vco_in * plln + 47_999_999) / 48_000_000;
+            pll48clk = Some(Hertz(vco_in * plln / pllq));
+
+            unsafe { &*RCC::ptr() }.pllcfgr.write(|w| unsafe {
```
Disasm, line 349:
```suggestion
            rcc.pllcfgr.write(|w| unsafe {
```

src/rcc.rs
```diff
@@ -286,91 +293,75 @@ impl CFGR {
...
+                    let plln = target_freq / vco_in;
+                    target_freq - vco_in * plln
+                })
+                .unwrap();
+
+            let vco_in = sysclk / pllm;
+            assert!(vco_in >= 1_000_000 && vco_in <= 2_000_000);
+
+            // Main scaler, must result in >= 100MHz (>= 192MHz for F401)
+            // and <= 432MHz, min 50, max 432
```
rfuest, line 327: Can you update this comment with the correct values for the F7 series?

src/rcc.rs
```diff
@@ -286,91 +293,75 @@ impl CFGR {
...
+            let pllm = (pllm_min..=pllm_max)
+                .min_by_key(|pllm| {
+                    let vco_in = sysclk / pllm;
+                    let plln = target_freq / vco_in;
+                    target_freq - vco_in * plln
+                })
+                .unwrap();
+
+            let vco_in = sysclk / pllm;
+            assert!(vco_in >= 1_000_000 && vco_in <= 2_000_000);
```
rfuest, line 324: According to the datasheet the valid input frequency range for the PLLs is 0.95 to 2.10 MHz.

no111u3: The software has to set these bits correctly to ensure that the VCO input frequency
ranges from 1 to 2 MHz. It is recommended to select a frequency of 2 MHz to limit
PLL jitter.

src/rcc.rs
no111u3: Done

rfuest:
Do we assume that the RCC peripheral is always in the reset state when its called? Otherwise the PLL code should also clear `DCKCFGR2.CK48MSEL` to make sure the correct PLL is used.

no111u3:
> Do we assume that the RCC peripheral is always in the reset state when its called? Otherwise the PLL code should also clear `DCKCFGR2.CK48MSEL` to make sure the correct PLL is used.

I think yes we always assume that the RCC freeze function call after reset state because as for other hal implementations we have some biggest changes for initialize another peripheral from its both state.

rfuest:
But I guess it wouldn't hurt to make sure the correct source is selected. Just to make sure no one has messed with the registers before the RCC is initialized.

no111u3:
> But I guess it wouldn't hurt to make sure the correct source is selected. Just to make sure no one has messed with the registers before the RCC is initialized.

Yep, It might be better, I try to do it.

mvertescher:
@Disasm, @rfuest any other thoughts on this change? I'll merge this in the next day or so if not...

Disasm:
I found that this change removes a lot of useful checks that I improve in https://github.com/stm32-rs/stm32f7xx-hal/pull/63, so I'd rather be careful with merging this as is.

no111u3:
> I found that this change removes a lot of useful checks that I improve in #63, so I'd rather be careful with merging this as is.

stm32f4 doesn't have this code too - two ways are incompatible in details.

Disasm:
@no111u3 it's a problem of `stm32f4xx-hal`. Let's make a compatible *public* interface and preserve the better clock setup implementation from `stm32f7xx-hal`.

Disasm:
I tested this code on STM32F723 (stlink-v3set board) with a simple test (https://gist.github.com/Disasm/8ede1b2676e949b27c19f9ccb17bb4f8) and found that it doesn't produce valid clock configurations for the following setups:
* HSE: 25MHz, bypass, SYSCLK: 48MHz, `require_pll48clk()`
* HSE: 12MHz, bypass, SYSCLK: 48MHz, `require_pll48clk()`

At the same time, master branch (500dae25b6a6ed1f5956074e830dfae2320b7ca6) produces valid results (apart from PLLQ) for these configurations.

no111u3:
Needs of totally rework with some architecture updates.

PR #60, opened 2020-07-04
Add USB driver for OTG_FS

Disasm:
Depends on https://github.com/stm32-rs/stm32f7xx-hal/pull/74

### DIFF
--- a/examples/usb_serial.rs
+++ b/examples/usb_serial.rs
@@ -6,15 +6,14 @@
 use panic_semihosting as _;
 
 use cortex_m_rt::entry;
-use stm32f7xx_hal::prelude::*;
-use stm32f7xx_hal::pac;
-use stm32f7xx_hal::rcc::{HSEClock, HSEClockMode};
 #[cfg(feature = "usb_fs")]
 use stm32f7xx_hal::otg_fs::{UsbBus, USB};
 #[cfg(feature = "usb_hs")]
 use stm32f7xx_hal::otg_hs::{UsbBus, USB};
+use stm32f7xx_hal::pac;
+use stm32f7xx_hal::prelude::*;
+use stm32f7xx_hal::rcc::{HSEClock, HSEClockMode};
 use usb_device::prelude::*;
-
 
 #[entry]
 fn main() -> ! {
--- a/examples/usb_serial.rs
+++ b/examples/usb_serial.rs
@@ -1,0 +1,98 @@
+//! CDC-ACM serial port example using polling in a busy loop.
+//! Target board: any STM32F7 with a OTG FS peripheral and a 25MHz HSE generator
+#![no_std]
+#![no_main]
+
+use panic_semihosting as _;
+
+use cortex_m_rt::entry;
+use stm32f7xx_hal::prelude::*;
+use stm32f7xx_hal::pac;
+use stm32f7xx_hal::rcc::{HSEClock, HSEClockMode};
+#[cfg(feature = "usb_fs")]
+use stm32f7xx_hal::otg_fs::{UsbBus, USB};
+#[cfg(feature = "usb_hs")]
+use stm32f7xx_hal::otg_hs::{UsbBus, USB};
+use usb_device::prelude::*;
+
+
+#[entry]
+fn main() -> ! {
+    let dp = pac::Peripherals::take().unwrap();
+
+    let rcc = dp.RCC.constrain();
+
+    let clocks = rcc
+        .cfgr
+        .hse(HSEClock::new(25.mhz(), HSEClockMode::Bypass))
+        .use_pll()
+        .use_pll48clk()
+        .sysclk(216.mhz())
+        .freeze();
+
+    #[cfg(feature = "usb_fs")]
+    let gpioa = dp.GPIOA.split();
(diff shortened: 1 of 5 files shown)


Disasm:
WTF, may I ask? This compiles perfectly fine on my local machine.

src/otg_fs.rs
```diff
@@ -0,0 +1,50 @@
+//! USB OTG full-speed peripheral
+//!
+//! Requires the `usb_fs` feature.
+//! Only one of the `usb_fs`/`usb_hs` features can be selected at the same time.
+
+use crate::device as stm32;
```
rfuest, line 6:
```suggestion
use crate::pac as stm32;
```

Disasm:
Rebased

Disasm:
I tested this driver (with a few clock setup hacks) on STM32F723, I assume that other chips documented in the same Reference Manual are supported too.

Disasm:
Since https://github.com/stm32-rs/stm32f7xx-hal/pull/74 has landed and now we have support for 48MHz USB clock setup, this PR is ready to be reviewed/merged.
I tested the usb_serial example in this PR on my F723-based board and it works.

Disasm:
@rfuest, @mvertescher could you review this PR?

src/otg_fs.rs
```diff
@@ -0,0 +1,53 @@
...
+use crate::gpio::{
+    gpioa::{PA11, PA12},
+    Alternate, AF10,
+};
+use crate::time::Hertz;
+
+pub use synopsys_usb_otg::UsbBus;
+use synopsys_usb_otg::UsbPeripheral;
+
+pub struct USB {
```
rfuest, line 17: Other peripherals, like Spi and Serial also provide a `new` constructor, IMO it would be more consistent to also add one for USB.

examples/usb_serial.rs
```diff
@@ -0,0 +1,97 @@
...
+#![no_std]
+#![no_main]
+
+use panic_semihosting as _;
+
+use cortex_m_rt::entry;
+#[cfg(feature = "usb_fs")]
+use stm32f7xx_hal::otg_fs::{UsbBus, USB};
+#[cfg(feature = "usb_hs")]
+use stm32f7xx_hal::otg_hs::{UsbBus, USB};
```
rfuest, line 12: Do you also plan to add `usb_hs` support in a future PR? Otherwise this should probably be removed.

rfuest (commented):
I couldn't test this in hardware because my STM32F730 board doesn't have a USB connector. But the code looks good overall and I only have two comments.

src/otg_fs.rs
Disasm: Good idea, I've added the constuctor.

examples/usb_serial.rs
Disasm: Yes, I'm working on this part and I'm planning to add HS support in the future.

src/otg_fs.rs
```diff
@@ -0,0 +1,77 @@
...
+    ///
+    /// Call `UsbBus::new` to construct and initialize the USB peripheral driver.
+    pub fn new(
+        usb_global: pac::OTG_FS_GLOBAL,
+        usb_device: pac::OTG_FS_DEVICE,
+        usb_pwrclk: pac::OTG_FS_PWRCLK,
+        pin_dm: PA11<Alternate<AF10>>,
+        pin_dp: PA12<Alternate<AF10>>,
+        clocks: Clocks,
+    ) -> Self {
```
rfuest, line 38: IMO the pins should be specified using a tuple to be consistent with the other peripherals. But I don't think there is any benefit in adding a `Pins` trait, like in the other peripherals, if there is only one valid pair of pins.
```suggestion
    pub fn new(
        usb_global: pac::OTG_FS_GLOBAL,
        usb_device: pac::OTG_FS_DEVICE,
        usb_pwrclk: pac::OTG_FS_PWRCLK,
        pins: (PA11<Alternate<AF10>>, PA12<Alternate<AF10>>),
        clocks: Clocks,
    ) -> Self {
```
or perhaps it's better to define a type for the pins:
```rust
pub type USB_DM = PA11<Alternate<AF10>>;
pub type USB_DP = PA12<Alternate<AF10>>;
// maybe: pub type USB_PINS = (USB_DM, USB_DP);
```

rfuest (changes requested):
I like the added constructor but have one nit about the consistency with the other peripherals.

src/otg_fs.rs
Disasm: I don't like neither of these options as they do not tell which pin is D+ and which is D-. What do you think about adding a type `struct Pins { dp: ..., dm: ...}` and passing it instead of a tuple?

rfuest: I agree that this isn't perfect, but I would prefer to be consistent with the other peripherals (in this and other STM32 HAL crates). The order of RX and TX for `Serial::new` can also be mixed up easily, but the compiler will complain if you get the order wrong.

Disasm: Got it, fixed it :)

mvertescher (approved):
Awesome, lgtm!

Issue #58, opened 2020-07-03
Create generation memory.x for different mcu's in family 


no111u3:
Than I tried build example for stm32fdisco-scree I got some interested message of compilation:
``` cargo build --example stm32f7disco-screen  --features="stm32f746 rt"
   Compiling stm32f7xx-hal v0.2.0 (/Users/nis/Develop/stm32f7xx-hal)
error: linking with `rust-lld` failed: exit code: 1
  |
  <skipped by unnecessary> 
  = note: rust-lld: error: section '.bss' will not fit in region 'RAM': overflowed by 130048 bytes
          rust-lld: error: section '.bss' will not fit in region 'RAM': overflowed by 130049 bytes
          rust-lld: error: section '.bss' will not fit in region 'RAM': overflowed by 130050 bytes
          rust-lld: error: section '.bss' will not fit in region 'RAM': overflowed by 130052 bytes
          rust-lld: error: section '.uninit' will not fit in region 'RAM': overflowed by 130052 bytes
          rust-lld: error: section '.uninit' will not fit in region 'RAM': overflowed by 130052 bytes
          rust-lld: error: section '.uninit' will not fit in region 'RAM': overflowed by 130052 bytes


error: aborting due to previous error

error: could not compile `stm32f7xx-hal`.
```

I think we need to use generation for memory.x

hannobraun:
There are multiple ways to do this, but here's some example code that could be copied from:
https://github.com/lpc-rs/lpc8xx-hal/blob/master/build.rs#L42-L63

no111u3:
> There are multiple ways to do this, but here's some example code that could be copied from:
> https://github.com/lpc-rs/lpc8xx-hal/blob/59b2a9d996570ad1869e172f8d4fed67958ce4c3/build.rs#L42-L63

Yes, sure. I talked about this solution - so If anybody hasn't concerns to this solution I can create PR for tests.

mvertescher:
Yeah, this is a great idea. Adopting the `lpc8xx-hal` implementation sounds good to me.

no111u3:
@mvertescher @hannobraun I think this issue can be close by success resolve.

PR #61, opened 2020-07-05
Add correct memory sizes selection for linker script

no111u3:
Following by issue #58 I tried to use `build.rs` for select current memory layout.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,145 @@
+use std::{
+    env,
+    fs::File,
+    io::{self, prelude::*},
+    path::PathBuf,
+};
+
+fn main() -> Result<(), Error> {
+    let target = Target::read();
+
+    copy_memory_config(target)?;
+
+    println!("cargo:rerun-if-changed=build.rs");
+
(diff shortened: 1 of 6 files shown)


build.rs
```diff
@@ -0,0 +1,145 @@
...
+        } else if cfg!(feature = "stm32f777") {
+            SubFamily::Stm32f777
+        } else if cfg!(feature = "stm32f778") {
+            SubFamily::Stm32f778
+        } else if cfg!(feature = "stm32f779") {
+            SubFamily::Stm32f779
+        } else {
+            error("You must select a target.
+If you added Stm32f7xx HAL as a dependency to your crate, you can select a target by enabling the respective feature in `Cargo.toml`.
+If you're running an example from the repository, select a target by passing the desired target as a command-line argument, for example `--features=824m201jhi33`.
```
mvertescher, line 118: Should probably change `824m201jhi33` to an f7 feature like `stm32f746`

no111u3: Sure, Done

PR #63, opened 2020-07-08
Fix clock setup code

Disasm:
According to the STM32F723ZE datasheet, minimal VCO output frequency is 100MHz (not 192MHz).
According to the RM0431 reference manual, 50 ≤ PLLN ≤ 432.

### DIFF
--- a/src/rcc.rs
+++ b/src/rcc.rs
@@ -271,14 +271,12 @@ impl CFGR {
         // Switch to HSI
         rcc.cfgr.modify(|_, w| w.sw().hsi());
 
-        // If HSE is provided by the user
-        let hse_freq: u32 = self.hse.as_ref().map_or(0, |c| c.freq);
-        // SYSCLK, must be <= 216 Mhz. By default, HSI frequency is chosen
-        let mut sysclk = self.sysclk.unwrap_or(HSI);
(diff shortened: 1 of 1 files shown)


Disasm:
For the case, when there is no override for SYSCLK and PLL is not used, SYSCLK must be equal to HSE or HSI frequency. However, in this case `sysclk` defaults to `HSI` even if HSE is used, which leads to the wrong `sysclk` value in `Clocks` structure when HSE frequency is different from 16MHz. The last commit improves consistency between `base_clk` and `sysclk` values.

src/rcc.rs
```diff
@@ -295,9 +301,9 @@ impl CFGR {
             let vco_clkin_mhz = (base_clk as f32 / pllm as f32) / 1_000_000.0;
             let mut sysclk_mhz: f32 = sysclk as f32 / 1_000_000.0;
 
-            // PLLN, main scaler, must result in >= 192MHz and <= 432MHz, min
-            // 50, max 432, this constraint is allways respected when vco_clkin
-            // <= 2 MHz
+            // PLLN, main scaler, must result in VCO frequency >=100MHz and <=432MHz,
```
mvertescher, line 304: Yep, this looks correct to me based on the reference manuals.

Issue #68, opened 2020-08-08
Add EXTI Example and Provide RCC Support for SYSCFG Clock


bbrown1867:
It would be useful to have an example showing how to use the EXTI traits for a GPIO input pin. 

While writing this example, I also noticed that `ExtiPin::make_interrupt_source` will not work unless we configure the SYSCFG bit of `apb2enr` in the RCC. It would be nice if this bit was enabled by default when we call`make_interrupt_source`, however since this is a larger "system level" configuration maybe it does not belong in `gpio.rs`? A simple solution may to just make `APB2::enr` in `rcc.rs` public to the user instead of just public to the crate.

I'll open a pull request to make this more clear.

mvertescher:
This issue should be resolved now, thanks again @bbrown1867!

PR #69, opened 2020-08-08
Add EXTI Example and Provide RCC Support for SYSCFG Clock

bbrown1867:
This pull request implements the example portion of #68 

The unsafe register write on lines 30-32 would ideally be replaced by something in the HAL. However since I'm a bit new to this crate I'm unsure what the best approach is (support in `gpio.rs`, `rcc.rs`?).

Build this example with the following commands:
`cargo build --example exti_button --features="stm32f767 rt"`

### DIFF
--- a/examples/exti_button.rs
+++ b/examples/exti_button.rs
@@ -1,0 +1,77 @@
+/// This example demonstrates how to use the ExtiPin trait for a GPIO input pin to capture a
+/// push button press. This example was written for the Nucleo-F767ZI board, from the Nucleo-144
+/// family, targeting the STM32F767ZI microcontroller. To port this to another board, simply
+/// change the GPIOs used for the push button and for the debug LED. Note that the EXTI number
+/// may change if using a new button, meaning that the interrupt handler will need to change also.
+///
+/// The intended behavior of the example is that when the user presses the button, an LED is
+/// toggled.
+
+#![no_main]
+#![no_std]
+
+extern crate panic_halt;
+
+use cortex_m::peripheral::NVIC;
+use cortex_m_rt::entry;
+use nucleof767zi_rs::Led;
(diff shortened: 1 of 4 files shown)


bbrown1867:
Done making changes to this PR now. I think the example is as safe as I can get it, and I've added the necessary register writes in `gpio.rs`.

src/gpio.rs
```diff
@@ -199,7 +199,10 @@ macro_rules! gpio {
 
             impl<MODE> ExtiPin for $PXx<Input<MODE>> {
                 /// Make corresponding EXTI line sensitive to this pin
-                fn make_interrupt_source(&mut self, syscfg: &mut SYSCFG) {
+                fn make_interrupt_source(&mut self, syscfg: &mut SYSCFG, rcc: &mut RCC) {
+                    // SYSCFG clock must be enabled in order to do register writes
+                    rcc.apb2enr.modify(|_, w| w.syscfgen().set_bit());
```
mvertescher, line 204: > However since I'm a bit new to this crate I'm unsure what the best approach is (support in gpio.rs, rcc.rs?).

This looks good to me atm despite it being a slightly different api than the F4 hal for instance. From the reference manual, the SYSCFG is only really used for a few things on the F7 (remap memory, select ethernet phy, exti). I could see maybe moving this to `rcc.rs` when more features are enabled, but seems ok here for now.

examples/exti_button.rs
```diff
@@ -0,0 +1,91 @@
+//! This example demonstrates how to use the ExtiPin trait for a GPIO input pin to capture a
```
mvertescher, line 1: :+1: , very helpful for folks learning how to use this crate

mvertescher:
Hey @bbrown1867, sorry for the delayed review on this. Looks good, thanks for this example!

PR #70, opened 2020-08-09
Add FMC via stm32-fmc crate

richardeoin:
Uses [stm32-fmc](https://github.com/stm32-rs/stm32-fmc)

* Only SDRAM supported so far
* stm32-fmc [only supports some of the f7 family](https://github.com/stm32-rs/stm32-fmc/blob/master/src/ral/peripherals/fmc.rs#L7), but I think it could be extended to cover the rest quite easily

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ nb = "0.1.2"
 stm32f7 = "0.11.0"
 micromath = "1.0.0"
 synopsys-usb-otg = { version = "0.2.0", features = ["cortex-m"], optional = true }
-stm32-fmc = { version = "0.1.2", features = ["sdram"], optional = true }
+stm32-fmc = { version = "0.2.0", features = ["sdram"], optional = true }
 
 [dependencies.bare-metal]
 version = "0.2.4"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ nb = "0.1.2"
 stm32f7 = "0.11.0"
 micromath = "1.0.0"
 synopsys-usb-otg = { version = "0.2.0", features = ["cortex-m"], optional = true }
+stm32-fmc = { version = "0.1.2", features = ["sdram"], optional = true }
 
 [dependencies.bare-metal]
 version = "0.2.4"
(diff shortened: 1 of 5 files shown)


examples/fmc.rs
```diff
@@ -0,0 +1,100 @@
+//! Initialises FMC controller
```
bbrown1867, line 1: Could you state what board and SDRAM chip were used for this example? It looks like is42s32800g? Did you use a breadboard to connect it to the STM32F7 or is there a special shield/board used?

richardeoin: I haven't tried this with a F7 yet, but indeed the example is written assuming a [IS42S32800G](http://www.issi.com/WW/pdf/42-45S32800G.pdf) SDRAM chip. There's several development boards from ST that have SDRAM, for instance the [STM32F769IDISCOVERY](https://www.st.com/en/evaluation-tools/32f769idiscovery.html) or the [STM32F746GDISCOVERY](https://www.st.com/content/st_com/en/products/evaluation-tools/product-evaluation-tools/mcu-mpu-eval-tools/stm32-mcu-mpu-eval-tools/stm32-discovery-kits/32f746gdiscovery.html). If someone was to test with one of those, then that would be an even better example!

Practically the SDRAM would always be on the same PCB as the F7, due to the high rise/fall times required. It might be possible over a high speed connector (Samtec QSH for example) if designed correctly, but I've never seen such a board.

ctron:
I would too be interested in this PR … trying to implement the display of the STM32723EDISCOVERY, which is attached via the FMC.

richardeoin:
@ctron Please go-ahead and try it out if you can. Good reports of it working will help the maintainers merge it. If you get stuck come ask in the stm32-rs channel.

However, I think [stm32-fmc](https://github.com/stm32-rs/stm32-fmc) crate will need some work before you can use it with the STM32F723EDISCOVERY. stm32-fmc only supports SDRAM so far, whilst that board has a Display and PSRAM.

examples/fmc.rs
bbrown1867: I have this example running on an STM32F746G Discovery Board. The board does not have the [IS42S32800G](http://www.issi.com/WW/pdf/42-45S32800G.pdf), but instead has the [IS42S32400F)[http://www.issi.com/WW/pdf/42-45S32400F.pdf] - at quick glance this seems to be the same part but half the size (128 MB instead of 256 MB). 

Is there any place I need to encode that size information in the FMC driver implementation for this chip? Right now I am just using the `is42s32800g_6` implementation with no changes. I don't see any place where 256 MB is called out for that chip. Maybe the driver doesn't care, and the application just makes sure we don't go over the available memory space?

Also the size in this example is 32 MB not 256 MB. Is that intentional? (1/8 internal SDRAM banks?)
`let ram_size_bytes = 32 * 1024 * 1024;`

bbrown1867: Scratch that last comment. I was confusing "Mb" and "MB". The device is 256 Mb so 32 MB makes sense.

Other observations I am seeing running this example with the IS42S32400F:
* I can only write/read on 4-byte boundaries (word aligned)
* Only the lower 16-bits of each word are accessible - the upper 16-bits stay 0. I think this is because on the STM32F746G Discovery Board only the lower 16 data pins are connected. This is a bit disappointing though, I was hoping I could write 32-bits to address X and not have to worry about that.

Cargo.toml
```diff
@@ -22,6 +22,7 @@ nb = "0.1.2"
 stm32f7 = "0.11.0"
 micromath = "1.0.0"
 synopsys-usb-otg = { version = "0.2.0", features = ["cortex-m"], optional = true }
+stm32-fmc = { version = "0.1.2", features = ["sdram"], optional = true }
```
bbrown1867, line 25: I get a compiler error when attempting to use version 0.2.0:

```
error[E0107]: wrong number of type arguments: expected 2, found 1
  --> /Users/ben/Documents/projects/stm32f7xx-hal/src/fmc.rs:75:39
   |
75 |     fn sdram<BANK: SdramPinSet, PINS: PinsSdram<BANK>, CHIP: SdramChip>(
   |                                       ^^^^^^^^^^^^^^^ expected 2 type arguments

error: aborting due to previous error
```

Might be good to have this crate work with the latest  version prior to merging.

examples/fmc.rs
richardeoin: Hi @bbrown1867, thanks for trying this out!

The size of an SDRAM device is the sum of the [rows, columns and number of banks](https://github.com/stm32-rs/stm32-fmc/blob/master/src/devices/is42s32800g.rs#L47). So for the IS42S32800 it's 2**(12+9) * 4 = 0x800_000 words (that's the "800" in the part name). Multiplying that by the bus width gives you the size in bits.

However those values are just used to configure the controller, instead the call to `init()` returns a raw pointer to the beginning of the memory-mapped block. It's then up to the application how much memory is actually used. I think that approach provides maximum flexibility to users who want a single slice, multiple slices or want to allocate the FMC memory at compile time.

Since the STM32F746G Discovery Board only has 16-bits of DQ connected, you should use a `FmcSdramConfiguration` with `memory_data_width` = 16. From the existing memory device definitions the [is42s16400j](https://github.com/stm32-rs/stm32-fmc/blob/master/src/devices/is42s16400j.rs) would be worth trying, especially since the number of words also matches the IS42S32400F. It might also affect the word alignment problem, since DQM[23] are also unconnected on STM32F746G Discovery Board.

bbrown1867: Thanks for the reply! Changing the `memory_data_width` did fix the problem I was having with only the lower 16-bits, I can now write to all 32-bits of each word.

bbrown1867:
@mvertescher is there anything preventing us from merging this into `master`? It worked on the STM32F746G Discovery Board when I tested it about a month ago.

mvertescher (approved):

mvertescher:
Nope, looks good. Glad to hear it works on one of the discovery boards. Thanks for this feature!

PR #74, opened 2020-08-27
Add new implementation for RCC

no111u3:
Create new version of RCC setup, with direct setup clocks.
Also update clock calculation and add usb pll configuration.

### DIFF
--- a/src/rcc.rs
+++ b/src/rcc.rs
@@ -2,7 +2,7 @@ use core::cmp::min;
 
 use micromath::F32Ext;
 
-use crate::pac::{rcc, FLASH, RCC};
+use crate::pac::{rcc, FLASH, PWR, RCC};
 use crate::time::Hertz;
 
 /// Extension trait that constrains the `RCC` peripheral
@@ -175,6 +175,19 @@ pub enum PLLP {
 }
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum VOSscale {
+    PwrScale1,
+    PwrScale2,
+    PwrScale3,
+}
+
+impl Default for VOSscale {
+    fn default() -> Self {
+        VOSscale::PwrScale3
+    }
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub struct CFGR {
     hse: Option<HSEClock>,
     use_pll: bool,
@@ -187,10 +200,12 @@ pub struct CFGR {
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Default)]
 struct InternalRCCConfig {
-    pub hpre: u8,
-    pub ppre1: u8,
-    pub ppre2: u8,
-    pub flash_waitstates: u8,
+    hpre: u8,
+    ppre1: u8,
+    ppre2: u8,
+    flash_waitstates: u8,
+    overdrive: bool,
+    vos_scale: VOSscale,
 }
 
 impl CFGR {
@@ -389,6 +404,15 @@ impl CFGR {
         } else {
             0b0111
         };
+        // Adjust power state and overdrive mode
+        config.vos_scale = if sysclk <= 144_000_000 {
+            VOSscale::PwrScale3
+        } else if sysclk <= 168_000_000 {
+            VOSscale::PwrScale2
+        } else {
+            VOSscale::PwrScale1
+        };
+        config.overdrive = if sysclk <= 180_000_000 { false } else { true };
 
         let clocks = Clocks {
             hclk: Hertz(hclk),
@@ -440,6 +464,7 @@ impl CFGR {
     pub fn freeze(self) -> Clocks {
         let flash = unsafe { &(*FLASH::ptr()) };
         let rcc = unsafe { &(*RCC::ptr()) };
+        let pwr = unsafe { &(*PWR::ptr()) };
 
         let (clocks, config) = self.calculate_clocks();
 
@@ -472,8 +497,22 @@ impl CFGR {
                 w.pllsrc().bit(self.hse.is_some())
             });
 
+            // Enable PWR domain and setup VOSscale and Overdrive options
+            rcc.apb1enr.modify(|_, w| w.pwren().set_bit());
+
+            pwr.cr1.modify(|_, w| match config.vos_scale {
+                VOSscale::PwrScale3 => w.vos().scale3(),
+                VOSscale::PwrScale2 => w.vos().scale2(),
+                VOSscale::PwrScale1 => w.vos().scale1(),
+            });
+
+            if config.overdrive {
+                pwr.cr1.modify(|_, w| w.oden().set_bit());
+            }
+
             // Enable PLL
             rcc.cr.modify(|_, w| w.pllon().on());
+
             // // Wait for PLL to stabilise
             while rcc.cr.read().pllrdy().is_not_ready() {}
         }
--- a/src/rcc.rs
+++ b/src/rcc.rs
@@ -21,6 +21,9 @@ impl RccExt for RCC {
             apb2: APB2 { _0: () },
             cfgr: CFGR {
                 hse: None,
+                hclk: None,
+                pclk1: None,
+                pclk2: None,
                 use_pll: false,
                 use_pll48clk: false,
                 pllm: 2,
@@ -190,6 +193,9 @@ impl Default for VOSscale {
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub struct CFGR {
     hse: Option<HSEClock>,
+    hclk: Option<u32>,
+    pclk1: Option<u32>,
+    pclk2: Option<u32>,
     use_pll: bool,
     use_pll48clk: bool,
     pllm: u8,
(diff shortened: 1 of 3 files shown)


src/rcc.rs
```diff
@@ -164,231 +166,117 @@ impl HSEClock {
 
 const HSI: u32 = 16_000_000; // Hz
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum PLLP {
+    Pllp2 = 0b00,
+    Pllp4 = 0b01,
+    Pllp6 = 0b10,
+    Pllp8 = 0b11,
```
Disasm, line 174: It's better to call them `Div2`, `Div4`, ...

src/rcc.rs
```diff
@@ -164,231 +166,117 @@ impl HSEClock {
...
+                    PLLP::Pllp2 => 2,
+                    PLLP::Pllp4 => 4,
+                    PLLP::Pllp6 => 6,
+                    PLLP::Pllp8 => 8,
+                };
 
-            use_pll = true;
+            if self.use_pll48clk {
+                pll48clk_valid =
+                    base_clk / self.pllm as u32 * self.plln as u32 / self.pllq as u32 == 48_000_000;
```
Disasm, line 269: According to the f4 hal code, USB specification allows +-0.25%

src/rcc.rs
```diff
@@ -164,231 +166,117 @@ impl HSEClock {
...
-
-        // Switch to fail-safe clock settings.
-        // This is useful when booting from a bootloader that alters clock tree configuration.
-        // Turn on HSI
-        rcc.cr.modify(|_, w| w.hsion().set_bit());
-        while rcc.cr.read().hsirdy().bit_is_clear() {}
-        // Switch to HSI
-        rcc.cfgr.modify(|_, w| w.sw().hsi());
+    /// Output clock calculation
+    fn calculation_clocks(&self) -> (Clocks, InternalRCCConfig) {
```
Disasm, line 246:
```suggestion
    fn calculate_clocks(&self) -> (Clocks, InternalRCCConfig) {
```

src/rcc.rs
```diff
@@ -164,231 +166,117 @@ impl HSEClock {
...
-                });
-            }
+        let mut pll48clk_valid = false;
 
-            // Enable PLL
-            rcc.cr.modify(|_, w| w.pllon().on());
-            // // Wait for PLL to stabilise
-            while rcc.cr.read().pllrdy().is_not_ready() {}
+        if self.use_pll {
+            sysclk = base_clk / self.pllm as u32 * self.plln as u32
```
Disasm, line 259:
```suggestion
            sysclk = base_clk * self.plln as u32 / self.pllm as u32
```

src/rcc.rs
no111u3: Done.

src/rcc.rs
no111u3: Done.

src/rcc.rs
```diff
@@ -478,28 +450,276 @@ impl CFGR {
...
+
+        Some((m, n, p, q))
+    }
+
+    /// Configure system clock settings by providen user clock frequency
+    /// Set sysclk as providen and setup USB clock if defined.
+    /// It must be between 12.5 Mhz and 216 Mhz.
+    /// If the ethernet peripheral is on, the user should set a
+    /// frequency higher than 25 Mhz
+    pub fn set_sysclk<F>(self, sysclk: F) -> Option<Self>
```
mvertescher, line 580: Maybe something like:

```
    /// Configure the system clock settings. 
    ///
    /// This sets the sysclk frequency and sets up the USB clock if defined.
    /// The provided frequency must be between 12.5 Mhz and 216 Mhz.
    /// If the ethernet peripheral is on, the user should set a frequency higher than 25 Mhz.
```

src/rcc.rs
```diff
@@ -478,28 +450,276 @@ impl CFGR {
...
+            });
+
+            if config.overdrive {
+                pwr.cr1.modify(|_, w| w.oden().set_bit());
+            }
+
+            // Enable PLL
+            rcc.cr.modify(|_, w| w.pllon().on());
+
+            // // Wait for PLL to stabilise
```
mvertescher, line 698: extra `//` here, maybe:  `// Wait for PLL to stabilise`

mvertescher:
We should do some testing on this PR before it lands just in case. If anyone has a chance to run through some sanity configurations that would be helpful. I may have some time later this week to test, unsure though.

src/rcc.rs
no111u3: Best variant, thanks.

src/rcc.rs
no111u3: Sure, removed.

src/rcc.rs
```diff
@@ -220,175 +269,85 @@ impl CFGR {
...
-            use_pll = true;
+        if self.use_pll48clk {
+            pll48clk_valid = {
+                let pll48clk =
+                    base_clk as u64 * self.plln as u64 / self.pllm as u64 / self.pllq as u64;
+                pll48clk >= 48_000_000 + 120_000 && pll48clk <= 48_000_000 + 120_000
+            };
         }
+        // SYSCLK, must be <= 216 Mhz. By default, HSI/HSE frequency is chosen
+        assert!(sysclk <= 216);
```
Disasm, line 343:
```suggestion
        assert!(sysclk <= 216_000_000);
```

Disasm:
Result of calculations depends on the position of `.set_sysclk()` call in a builder chain, this is not good.

src/rcc.rs
```diff
@@ -618,17 +618,15 @@ impl CFGR {
             if let Some(q) = q {
                 self.pllq(q as u8);
             }
-
-            Some(self)
-        } else {
-            None
         }
```
Disasm, line 621: else panic?

src/rcc.rs
no111u3: Done.

src/rcc.rs
no111u3: Done.

no111u3:
> Result of calculations depends on the position of `.set_sysclk()` call in a builder chain, this is not good.

Fixed.

no111u3:
> Also this test fails:
> 
> ```rust
>     #[test]
>     fn test_rcc_calc1() {
>         use super::{PLLP, HSEClock, HSEClockMode};
>         use crate::time::U32Ext;
> 
>         let cfgr = CFGR {
>             hse: None,
>             hclk: None,
>             pclk1: None,
>             pclk2: None,
>             use_pll: false,
>             use_pll48clk: false,
>             pllm: 2,
>             plln: 50,
>             pllp: PLLP::Div2,
>             pllq: 2,
>         };
> 
>         let cfgr = cfgr
>             .hse(HSEClock::new(25.mhz(), HSEClockMode::Bypass))
>             .use_pll()
>             .use_pll48clk()
>             .sysclk(216.mhz());
> 
>         let (clocks, config) = cfgr.calculate_clocks();
>         assert_eq!(clocks.sysclk().0, 216_000_000);
>         assert!(clocks.is_pll48clk_valid());
>     }
> ```

Fixed, test added to test group in rcc.

mvertescher:
@Disasm any other thoughts on this PR? If not, I'll merge this and we can follow up with any issues that arise.

src/rcc.rs
```diff
@@ -478,28 +470,272 @@ impl CFGR {
...
+    ///
+    /// The implementation makes the following choice: HSI is always chosen over
+    /// HSE except when HSE is provided. When HSE is provided, HSE is used
+    /// wherever it is possible.
+    pub fn freeze(self) -> Clocks {
+        let flash = unsafe { &(*FLASH::ptr()) };
+        let rcc = unsafe { &(*RCC::ptr()) };
+        let pwr = unsafe { &(*PWR::ptr()) };
+
+        self.pll_configure();
```
Disasm, line 664: `pll_configure` is Self->Self, so this operation is no-op.

no111u3: Fixed

src/rcc.rs
```diff
@@ -656,7 +654,7 @@ impl CFGR {
     /// The implementation makes the following choice: HSI is always chosen over
     /// HSE except when HSE is provided. When HSE is provided, HSE is used
     /// wherever it is possible.
-    pub fn freeze(self) -> Clocks {
+    pub fn freeze(&mut self) -> Clocks {
```
Disasm, line 657: I think it's a bad idea...

no111u3: Fixed.

Disasm (approved):
I tested 4c9af734cdb0b98cd60eecd873a038109a0f7c89 within this PR on my STLinkV3-SET as target board and both rcc and USB serial tests passed. Let's see what CI tells us...

Disasm:
Apparently, tests should be updated to reflect `pll_configure()` signature change.

no111u3:
> Apparently, tests should be updated to reflect `pll_configure()` signature change.

Fixed.

mvertescher (approved):

Issue #82, opened 2020-09-25
PLL isn't enabled automatically


rfuest:
Is the RCC HAL implementation supposed to enable the PLL automatically? I didn't have time to investigate myself, but I think it is: https://github.com/stm32-rs/stm32f7xx-hal/blob/caf2196fcfb10a2947b255f0437a4e9afe99000e/src/rcc.rs#L623

But if I try to configure the RCC using the following code the assertions fails and sysclk is 8 MHz:
```rust
let clocks = rcc
    .cfgr
    .hse(HSEClock::new(8.mhz(), HSEClockMode::Oscillator))
    .sysclk(216.mhz())
    .freeze();

assert_eq!(clocks.sysclk(), 216.mhz().into());
```

If I add a `.use_pll()` the configuration works as expected. @no111u3 could you clarify if this is a bug or the intended behavior?

pftbest:
This looks like a bug, the CFGR struct should not have #[derive(Copy)]

`use_pll` takes self by value making a new copy and returns it. The original CFGR is not changed.

no111u3:
The original CFGR doesn't change because its never needs to change. I fix configuration for PLL in #83. @rfuest thanks for report of issue.

rfuest:
Closed by #83.

PR #83, opened 2020-09-26
Change pll enable to direct

no111u3:
Fix the #82 issue.


rfuest (approved):
Thanks for the quick fix. I've tried your changes and everything works as expected now.

mvertescher (approved):
Thanks for the fix!

PR #88, opened 2020-10-21
Fix SRAM size for 45,46,56 devices

### LINKED ISSUE #81: Wrong RAM Size for STM32F7{45,46, 56} models?
The file `memory_1024_240.x` appears to declare the RAM section as 256KB for these models. However, I think the correct amount is 320KB.

From the table on [this page](https://www.st.com/en/microcontrollers-microprocessors/stm32f7-series.html):
table

Also from the datasheet for STM32F745xx/STM32F746xx:
datasheet

As far as I can tell, this change would effect two files:

- `memory_1024_240.x` 
- `build.rs`

Happy to make the fix if others can concur.

bbrown1867:
Closes #81 

Please see the screenshots on the issue for details.


hannobraun:
Thank you, @bbrown1867, looks good to me!

Issue #89, opened 2020-10-22
Migrate from Travis CI to Github Actions


mvertescher:
Currently, we use Travis CI for building, testing and formatting this crate. However, Travis CI is not as featured as the new Github Actions offering. Additionally, some of the other stm32-hal crates have already migrated. To keep this crate consistent and improve CI overall we should switch over.

PR #90, opened 2020-10-22
Initial Github Actions CI

mvertescher:
This patch provides an initial port of the current Travis CI
configuration to Github Actions to improve our CI overall and be more
consistent with the other stm32-hal crates. It also enables a basic
`clippy` lint check on the crate. At the moment, there are quite a few
warnings and errors that should be fixed in the future. Travis CI
support has been left unmodified for now, but should be removed later
too.

This is part 1 of #89

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,54 @@
+on: [push, pull_request]
+
+name: Continuous integration
+
+jobs:
+  ci:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        mcu:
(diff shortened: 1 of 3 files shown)


hannobraun:
Looks good to me! Not sure, if it should already run on this pull request, or if that requires the configuration to be in master. I guess we'll see soon.

hannobraun:
Everything runs fine on master. The only thing left to test is whether it runs on pull requests too. Good job, @mvertescher!

mvertescher:
Looks like it ran on #91, seems good to me :+1:

PR #91, opened 2020-10-22
Changes to FMC Example to Work on STM32F746G Discovery Board

bbrown1867:
As mentioned in #70, I was able to get the FMC example working on an F7 board - the example was originally written for an F4 board. I had to make some changes to the example though, using a different SDRAM device struct and changing GPIO pins. I've also improved the doc comment.


mvertescher (approved):
Awesome, looks great. Worked for me on my discovery board. Thanks for following up on this!

PR #94, opened 2020-10-25
Don't panic when PLL is not required

Disasm:


src/rcc.rs
```diff
@@ -1165,4 +1171,28 @@ mod tests {
         assert_eq!(clocks.sysclk().0, 216_000_000);
         assert!(clocks.is_pll48clk_valid());
     }
+
+    #[test]
+    fn test_rcc_default() {
```
mvertescher, line 1176: Nice :+1:

mvertescher (approved):
LGTM, thanks!

PR #95, opened 2020-10-28
WAITING FOR REVIEW: simple and fast way to get i2c clock, tested(for results check the comments)

### LINKED ISSUE #71: I2C timing calculation broken
Looks like the logic which calculates the value of the TIMINGR register is broken.

The code is located here: https://github.com/stm32-rs/stm32f7xx-hal/blob/master/src/i2c.rs#L338
One obvious problem with it, is that after:
`
     let fscll_mhz: f32 = base_clk_mhz / (scll as f32 + 1.0);
`
the prescaler is then calculated with
`
      let presc = base_clk_mhz / fscll_mhz;
`
which could be simplified to just
`
   let presc = sccl + 1;
`
which is obiously wrong.

That said, I could not decipher how the entire logic is even supposed to work...

I see two possible ways to fix the problem:
1. Instead of calculating this at runtime, have a way to provide this parameter as an argument to I2C initialization. TIMINGR can then be calculated using STM32CubeMX tool as can be seen in the screenshot below:

image

2. Improve the calculation logic. I found pretty complex (and probably slow) implementation in the STM32Duino probject:

https://github.com/stm32duino/Arduino_Core_STM32/blob/master/libraries/Wire/src/utility/twi.c#L383

7h3qu1rkyb1t:
faster algorithm to find timingr register values,
implemented a custom enum for custom timingr value,
sdadel is default made as 0, so no need of  vddat calculation, and hd:dat, still su:dat need to implement, i will add it soon,
took few concept from `craigjb:i2ctiming` this PR

<s> Needed some test results, i couldn't generate 1 Mhz clock, i dont know whats the problem, tried custom values given in datasheets,
reviews are requested and test results</s>

fixes  #71

### DIFF
--- a/src/i2c.rs
+++ b/src/i2c.rs
@@ -4,8 +4,6 @@
 // NB : this implementation started as a modified copy of https://github.com/stm32-rs/stm32f1xx-hal/blob/master/src/i2c.rs
 
 #[cfg_attr(test, allow(unused_imports))]
-use micromath::F32Ext;
-
 use crate::gpio::gpioa::PA8;
 use crate::gpio::gpiob::{PB10, PB11, PB6, PB7, PB8, PB9};
 use crate::gpio::gpioc::PC9;
@@ -16,6 +14,7 @@ use crate::hal::blocking::i2c::{Read, Write, WriteRead};
 use crate::pac::{DWT, I2C1, I2C2, I2C3};
 use crate::rcc::{sealed::RccBus, Clocks, Enable, GetBusFreq, Reset};
 use crate::time::Hertz;
+use micromath::F32Ext;
 use nb::Error::{Other, WouldBlock};
 use nb::{Error as NbError, Result as NbResult};
 
@@ -47,6 +46,7 @@ pub enum Mode {
     Standard { frequency: Hertz },
     Fast { frequency: Hertz },
     FastPlus { frequency: Hertz },
+    Custom { timing_r: u32 },
 }
 
 impl Mode {
@@ -65,14 +65,6 @@ impl Mode {
     pub fn fast_plus<F: Into<Hertz>>(frequency: F) -> Self {
         Mode::FastPlus {
             frequency: frequency.into(),
-        }
-    }
-
-    pub fn get_frequency(&self) -> Hertz {
-        match self {
-            &Mode::Standard { frequency } => frequency,
-            &Mode::Fast { frequency } => frequency,
-            &Mode::FastPlus { frequency } => frequency,
         }
     }
 }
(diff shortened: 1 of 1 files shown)


pftbest:
Which resistor values do you use for testing? The final frequency depends on the rise time which in turn depends on resistor values.

7h3qu1rkyb1t:
i was connecting it directly to my logic analyzer so didn't pick any resistor, do i need one?

pftbest:
Yes)

i2c lines are open drain so they can short to the ground but they can't charge back up to 3v without a pull-up resistor.

The lower the resistor value the faster it can charge parasitic capacitance so it will do a faster transition from 0 to 1.

Try 4.7k or 1k pull-up resistors.

7h3qu1rkyb1t:
i tested with with 2.2k resistor, but didn't worked fork me, currently i dont have cubemx setup do you know timingr register value for 1mhz, i tried for 0x50100103 for 48 mhz as given in datasheet but no use

pftbest:
For 216 MHz main with 54 MHz pclk1 cube gave me value 0x00200922 and I can measure around 960 kHz on logic analyzer.

For 48 MHz main with 48 MHz pclk1 the value is 0x20000209 and I get the same 960 kHz clock

my resistors are 1k each

pftbest:
When I tried your value 0x50100103 @ 48MHz I got exactly 1 MHz clock

Screen Shot 2020-10-28 at 19 42 23

7h3qu1rkyb1t:
ok, thank you for the results, its sdadel problem, i had hardcoded it to 4 for testing, i think that was causing problem.

7h3qu1rkyb1t:
done some calibrations, to get optimal frequencies without being heavy on compute. Here are some results
| i2c_clk  in mhz | expected scl in khz | practical scl in khz |
|---------------------|---------------------------|--------------------------|
| 16                    |  10                          |  9.95                      |
| 16                    |  100                        |  99.17                    | 
| 16                    |  400                        |  389                       |
| 16                    |  1000                      |  940                       |
| 32                    |  10                          |   error                    |
| 32                    |  100                        |   99.8                     |  
| 32                    |  400                        |  393                       |
| 32                    |  1000                      |  1000                     |
| 54                    |  10                          |   10.131                 |
| 54                    |  100                        |  99.767                  |   
| 54                    |  400                        | 412                        |
| 54                    |  1000                      | 1054                      |

having some problems for 10khz in between frequencies that will be addressed in next commit.
if anyone can please confirm this result.

7h3qu1rkyb1t:
removed get frequency method from mode enum as I couldn't find a way to calculate without i2cclk which is not available in mode enum.

7h3qu1rkyb1t:
Forgot to mention, i skipped calculation of busch of parameters, 
-  kept sda always 0, thats what cube mx also does
-  since sda is 0, vddat is always less than vddat max so no need to calculate vddat

7h3qu1rkyb1t:
Is there anything else required, or is it incorrect? Should i close this PR then?

src/i2c.rs
```diff
@@ -68,6 +69,15 @@ impl Mode {
             frequency: frequency.into(),
         }
     }
+
```
7h3qu1rkyb1t: Removed due to complexity of calculating frequency in case of custom, also it seems like unnecessory

7h3qu1rkyb1t:
Its seems like this PR is unnecessary, since no one ale reviewing it. I am closing this PR

hannobraun:
How does this relate to #72? Is it an alternative to it?

7h3qu1rkyb1t:
#72 using some brute force methods to calculate value, which is not required, and some of the calculations done in their was unnecessary.
It is kind of alternative,

Cargo.toml
```diff
@@ -23,6 +23,7 @@ stm32f7 = "0.11.0"
 micromath = "1.0.0"
 synopsys-usb-otg = { version = "0.2.0", features = ["cortex-m"], optional = true }
 stm32-fmc = { version = "0.2.0", features = ["sdram"], optional = true }
+rtt-target = {features = ["cortex-m"], version = "0.2.2"}
```
hannobraun, line 26: That doesn't seem to be used anywhere. Maybe remove it?

src/i2c.rs
```diff
@@ -15,6 +13,7 @@ use crate::hal::blocking::i2c::{Read, Write, WriteRead};
 use crate::pac::{DWT, I2C1, I2C2, I2C3};
 use crate::rcc::{sealed::RccBus, Clocks, Enable, GetBusFreq, Reset};
 use crate::time::Hertz;
+use micromath::F32Ext;
```
hannobraun, line 16: Not a big deal, but there doesn't seem to be any reason for moving this. If so, it would be better to leave this as is, so keep the change small.

src/i2c.rs
```diff
@@ -66,14 +66,6 @@ impl Mode {
...
         }
     }
-
-    pub fn get_frequency(&self) -> Hertz {
-        match self {
-            &Mode::Standard { frequency } => frequency,
-            &Mode::Fast { frequency } => frequency,
-            &Mode::FastPlus { frequency } => frequency,
-        }
-    }
```
hannobraun, line 76: I'm assuming this was removed, because a frequency is not available when using `Mode::Custom`. I'm not sure if we need it. If not, removing it is probably the best option.

If we do need it, it might be better to return `Option<Hertz>` instead.

src/i2c.rs
```diff
@@ -297,67 +411,52 @@ macro_rules! hal {
...
+                    i2c.init(clocks);
                     i2c
                 }
 
                 /// Initializes I2C as master. Configures I2C_PRESC, I2C_SCLDEL,
                 /// I2C_SDAEL, I2C_SCLH, I2C_SCLL
                 ///
                 /// For now, only standard mode is implemented
-                fn init(&mut self) {
+                fn init(&mut self, _clocks: Clocks) {
```
hannobraun, line 423: `_clocks` doesn't seem to be used. Why add it? To make sure that RCC config has been frozen?

hannobraun (commented):
Thank you, @7h3qu1rkyb1t! I left some questions and suggestions.

hannobraun:
Oh, also, this needs a rebase. Might be best to squash the commits while you're at it.

@craigjb, what do you think of merging this instead of #72? More efficient is better, of course, but other than that I'm not really qualified to judge the relative merits of both pull requests (and don't have the time to get qualified right now), so I'm hoping you can just hash this out among yourselves :-)

7h3qu1rkyb1t:
yeah @craigjb i too think it would be great if you check it once, i skipped some calculations based on some assumptions

src/i2c.rs
7h3qu1rkyb1t: its cargo format, changing order,

hannobraun: Alright then, whatever.

src/i2c.rs
7h3qu1rkyb1t: I feel like it is unnecessary, making things complicated. Frequency is a tuple struct so if anyone wants they can use .0 for that

7h3qu1rkyb1t: I feel like it is unnecessary, making things complicated. Frequency is a tuple struct so if anyone wants they can use .0 for that

hannobraun (commented):
This still needs a rebase (conflict in i2c.rs). I'm also still hoping that someone with more expertise will weigh in, but if not, I'm willing to merge it (after the conflict is fixed) if no one protests.

7h3qu1rkyb1t:
please let me know if i am doing anything wrong with git history, i am still learning to how to use git,

hannobraun:
> please let me know if i am doing anything wrong with git history, i am still learning to how to use git,

Hmm, I'm confused by that first commit, "Add F767 to features that support USB.". There's a commit with the same title in master, but [it looks completely different](https://github.com/stm32-rs/stm32f7xx-hal/commit/3b4fccc9c90d8496a6e32614efa8fc2d14106641). How did that get in there, and what does it do?

7h3qu1rkyb1t:
I think i missed it while squashing, hold on i will re do it again

hannobraun:
Looking good now! Thank you, @7h3qu1rkyb1t!

If you or others can confirm that it still works as expected after the rebase, and if no one objects, I'll merge.

7h3qu1rkyb1t:
i am getting same results,

7h3qu1rkyb1t:
i havent checked for all frequencies but i am currently working on a project which uses this PR, i am getting 390 KHz for 400KHz frequency

hannobraun:
Okay, I'm going to merge this now. If any problems still persist, I'm sure we'll notice at some point :-)

Thank you, @7h3qu1rkyb1t!

PR #97, opened 2020-11-05
Fix warnings in RCC tests

Disasm:
There are also a few warnings about unused `micromath::F32Ext` import when building tests, but I don't know why they happen and don't change these for now.


rfuest:
The `micromath` docs have a section about the `F32Ext` warning, if you want to fix them in this PR: https://docs.rs/micromath/1.1.0/micromath/#unused-import-warnings-when-linking-std

hannobraun:
Looks good to me, thank you @Disasm!

PR #98, opened 2020-11-05
Fix warnings

Disasm:
See also: https://github.com/stm32-rs/stm32f7xx-hal/pull/97#issuecomment-722434890


mvertescher (approved):
Awesome, thanks for fixing this up @Disasm!

PR #101, opened 2020-11-16
Add USB driver for OTG_HS

Disasm:
Based on the code from the [usb_midi](https://github.com/perlindgren/stm32f7xx-hal/tree/usb_midi) branch by @perlindgren.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,5 +50,9 @@ jobs:
           args: --features=${{ matrix.mcu }},rt,usb_fs --examples
       - uses: actions-rs/cargo@v1
         with:
+          command: build
+          args: --features=${{ matrix.mcu }},rt,usb_hs --examples
+      - uses: actions-rs/cargo@v1
+        with:
           command: test
           args: --features=${{ matrix.mcu }} --target x86_64-unknown-linux-gnu --lib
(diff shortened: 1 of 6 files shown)


perlindgren (approved):
Look excellent. 
```
    /// Call `UsbBus::new` to construct and initialize the USB peripheral driver.
    pub fn new(
```
Maybe it should be noted that the external Phy needs to be initialized.

Disasm:
As long as we usually force user to reconfigure all the required pins into appropriate AFs and pass them to driver, I think that external PHY case will be handled in a separate constructor function. I still have to figure out an interface for this.

Disasm:
Well, I tested this PR on a F723 Discovery board and everything works (apart from the usbd-serial compliance issue). I decided to leave the UPLI PHY support for now. I have a board with an external PHY, but it's a F407 board, so I'm going to add external PHY support to `stm32f4xx-hal` first and then maybe we port it to `stm32f7xx-hal`.

Ready for review.

Disasm:
Also do we want to make the `synopsys-usb-otg` dependency non-optional like it was done for `stm32-fmc`? It has a nice build time of 4-5s (most of which is `cortex-m` and its deps) and only 3 dependencies.

hannobraun:
As usual, I don't have the time for anything substantial, but I did a cursory review and this looks good. Merging, since @Disasm successfully tested it and this hasn't received any negative feedback.

Thank you, @Disasm!

Disasm:
@hannobraun thank you!

Issue #102, opened 2020-11-19
Implementation for RNG


7h3qu1rkyb1t:
stm32f4xx-hal has an implementation for RNG, Is this ok to Copy that or it will create any problems?

PR #103, opened 2020-11-24
[rdy] random number generation implimented

7h3qu1rkyb1t:
It is same implementation done in stm32f4xx-hal, but with a small change, i.e. they used software calculation for checking whether rng clk condition meets the requirements or not. which prone to errors as software calculation may not give exact frequency. 
i used status flags for that. 
this addresses #102

### DIFF
--- a/examples/rng.rs
+++ b/examples/rng.rs
@@ -1,0 +1,22 @@
+#![deny(unsafe_code)]
+#![deny(warnings)]
+#![no_main]
+#![no_std]
+
+/// generates a random number and displays it to the openocd console
+use panic_semihosting as _;
+
+use cortex_m_rt::entry;
+use cortex_m_semihosting::hprintln;
+use stm32f7xx_hal::{pac, prelude::*};
+
+#[entry]
+fn main() -> ! {
+    let p = pac::Peripherals::take().unwrap();
+    let mut rng = p.RNG.init();
+    let val = rng.get_rand().unwrap();
+    hprintln!("random value {}", val).unwrap();
+    loop {
+        core::sync::atomic::spin_loop_hint();
+    }
+}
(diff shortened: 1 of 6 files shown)


examples/rng.rs
```diff
@@ -8,20 +8,13 @@ use panic_semihosting as _;
...
 fn main() -> ! {
     let p = pac::Peripherals::take().unwrap();
-    let rcc = p.RCC.constrain();
-    let _clocks = rcc.cfgr.sysclk(32.mhz()).freeze();
     let mut rng = p.RNG.init();
     let val = rng.get_rand().unwrap();
     hprintln!("random value {}", val).unwrap();
-    loop{
-    }
+    loop {}
```
hannobraun, line 19: Empty loops can cause undefined behavior, due to a bug in LLVM (https://github.com/rust-lang/rust/issues/28728). I believe that adding `core::sync::atomic::spin_loop_hint();` to the body of the loop is the canonical way to prevent this.

```suggestion
    loop {
        core::sync::atomic::spin_loop_hint();
    }
```

hannobraun (changes requested):
Thank you, @7h3qu1rkyb1t!

I did a cursory review and found one small problem. There's also a conflict in Cargo.toml, so this needs a rebase.

examples/rng.rs
7h3qu1rkyb1t: Yeah i forgot to add it

src/prelude.rs
```diff
@@ -5,4 +5,5 @@ pub use crate::gpio::GpioExt as _stm327xx_hal_gpio_GpioExt;
 pub use crate::hal::digital::v2::{InputPin, OutputPin};
 pub use crate::hal::prelude::*;
 pub use crate::rcc::RccExt as _stm32f7xx_hal_rcc_RccExt;
+pub use crate::rng::RngExt as _stm32f7xx_hal_rng_RngExt;
```
hannobraun, line 8: This is no longer the best way to do a re-export (it used to be some time ago). Now this is better:
```suggestion
pub use crate::rng::RngExt as _;
```

hannobraun (changes requested):
I found another minor problem. Let's include this, then this is ready to be merged.

hannobraun (approved):
Turns out I can apply my own suggestions. Merging.

Thanks again, @7h3qu1rkyb1t!

Issue #105, opened 2020-12-16
LCD example shown black screen (with backlight on)


mchodzikiewicz:
I cloned repo and launched example with:
```
cargo +stable run --example stm32f7disco-screen --features="stm32f746 rt"
```

I have my discovery board plugged into USB, flashed succesfuly, when I halt the MCU in the gdb it shows it is in an infinite loop at the end of the example, but I cannot see any pixel drawn, just dark blank screen.

PS. I have tried it on two STM32F746G-DISCO boards and they both are fine with ST demos

PR #106, opened 2020-12-17
Fix LCD timings for F746G disco example

mchodzikiewicz:
This PR fixes https://github.com/stm32-rs/stm32f7xx-hal/issues/105

### DIFF
--- a/examples/stm32f7disco-screen/screen.rs
+++ b/examples/stm32f7disco-screen/screen.rs
@@ -20,14 +20,14 @@ pub const DISCO_SCREEN_CONFIG: DisplayConfig = DisplayConfig {
     active_height: 272,
     h_back_porch: 13,
     h_front_porch: 30,
-    h_sync: 30,
+    h_sync: 41,
     v_back_porch: 2,
     v_front_porch: 2,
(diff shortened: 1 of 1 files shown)


mchodzikiewicz:
btw, timings are adjusted according to official ST example: https://github.com/STMicroelectronics/STM32CubeF7/tree/master/Drivers/BSP/STM32746G-Discovery

hannobraun:
I haven't used this feature, nor do I have the hardware to test it. @grodino, it seems you added that code, any thoughts?

grodino:
@mchodzikiewicz The values I used were taken from an Application Note from ST and worked for me but if this is from the official example, I'm fine with this too ;)
@hannobraun I do not have the hardware anymore so I can't test it sorry

hannobraun:
Okay then, since it works for @mchodzikiewicz and nobody objects, I'm merging. Thank you @mchodzikiewicz and @grodino!

mchodzikiewicz:
great! What I can tell is that my two boards are from different hw revisions (B-02 and B-04) and both of them did not worked and started working after this adjustment

PR #108, opened 2020-12-26
CAN support

santiagoalvarez:


xoviat:
@santiagoalvarez Have you been able to get this working?

santiagoalvarez:
@xoviat not yet as I only have the STM32F746NG board available to me and I need to somehow get the code to work correctly in order to remap the pins since the default pins are not available

santiagoalvarez:
@xoviat it should at least allow the can to be enabled, I think the timing is still not correct though

santiagoalvarez:
@xoviat I have got it working on the disco board

hannobraun:
What's the status of this PR? I don't have time to test it, but I'm happy to merge it, if it works for you.

The only blocker is that CI is failing. The error looks unrelated, but still, it would be nice to see this passing before merge, as the unrelated error might mask other problems.

santiagoalvarez:
Ok, CI passes now and I can attest that this works fine (at least on my F7469NG-DISCO board)

hannobraun:
Thank you, @santiagoalvarez!

PR #110, opened 2021-01-05
Add STM32F746 to list of supported USB devices

andybest:
I have tested with the USB serial example and can confirm all works as expected


hannobraun:
Great, thank you @andybest!

PR #111, opened 2021-01-10
Use CamelCase for gpio::Edge enum variants

Piroro-hs:
ref: [h7xx](https://github.com/stm32-rs/stm32h7xx-hal/pull/179), [l4xx](https://github.com/stm32-rs/stm32l4xx-hal/pull/187), [f1xx](https://github.com/stm32-rs/stm32f1xx-hal/pull/303), [f4xx](https://github.com/stm32-rs/stm32f4xx-hal/pull/250)


hannobraun:
That makes a lot of sense. Thank you, @Piroro-hs!

PR #112, opened 2021-03-03
Adds the RCC bus data for SPI4 and SPI5

santiagoalvarez:


hannobraun:
Thank you, @santiagoalvarez!

PR #113, opened 2021-03-30
added dac support

hardfau18:
solves #9 
shamelessly copied and modified from stm32f4xx-hal
in future  will add alignment and different bit supports

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ micromath = "1.0.0"
 synopsys-usb-otg = { version = "0.2.3", features = ["cortex-m"], optional = true }
 stm32-fmc = { version = "0.2.0", features = ["sdram"], optional = true }
 rand_core = "0.5.1"
-bxcan = "0.4"
+bxcan = "0.5"
 
 [dependencies.bare-metal]
 version = "0.2.4"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -25,7 +25,7 @@ micromath = "1.0.0"
 synopsys-usb-otg = { version = "0.2.3", features = ["cortex-m"], optional = true }
 stm32-fmc = { version = "0.2.0", features = ["sdram"], optional = true }
 rand_core = "0.5.1"
-bxcan = "0.4"
```
hardfau18, line 28: defmt 2.0 requires this

hannobraun (commented):
Thank you, @hardfau18!

Some comments:
- I don't understand the reasoning for having a public `dac` function _and_ the `DacExt` trait. Seems redundant, but I guess it's fine. We can always clean it up later, whenever someone has the time and inclination to do so.
- I'm not a fan of the extension trait style (`DacExt`), as I think it introduces complexity (user has to import the trait) for no good reason. I prefer peripheral structs with a simple `new` constructor instead. But again, it's fine. Can always be cleaned up later.

I assume you tested this successfully? If so, I'd be happy to merge it.

hardfau18:
tested it.
About changes requested I am working on rewriting everything. but may take some time

hannobraun:
Great, thanks again!

Issue #114, opened 2021-04-01
bxcan/defmt  version conflict


pdgilbert:
In code that uses multiple device crates I have started having `cargo update` problems because of the change in the dependency  `bxcan = "0.5"` (see https://github.com/stm32-rs/bxcan/issues/22#issue-848677821). Can the dependency be more flexible, say `bxcan = ">=0.4"` ?

hannobraun:
I'm not familiar with `bxcan`, or the details of how we use it (I reviewed the PR adding it, but have since forgotten all about it). Does this HAL actually work with multiple versions of `bxcan`? Can we be sure that, even if it works for us, this can't cause breakage for our users?

If the answer to both questions is "yes", then I'm happy to merge a PR that relaxes the version requirement. Although it shouldn't be open-ended, but restricted to versions that are known to work (so `bxcan = ">=0.4, <0.6"` would probably be correct).

pdgilbert:
(Anyone please correct me if I say something is wrong. This is the first time I've run into this.)

The problem seems to be that `defmt` uses a `link`, so it is not possible for `bxcan` to specify a range of versions for `defmt`. A specific version of `bxcan` is tied to a specific version of `defmt`. (See https://github.com/knurling-rs/defmt/issues/426#issue-823703071 ) Thus `stm32f7xx-hal` specifying a specific version of `bxcan` propagates the  `defmt` conflict problem to anything linked with `stm32f7xx-hal`. 

>Does this HAL actually work with multiple versions of bxcan?

Technically I think this is not about "multiple versions" but rather about allowing a range of versions so that a single version can be selected when other linked crates also use `bxcan`.  `stm32f7xx-hal` CI seems to work when `bxcan = ">=0.4, <0.6"` is allowed.  My workflow testing at https://github.com/pdgilbert/eg_stm_hal/actions/runs/712196680 and at https://github.com/pdgilbert/LoRaGPS-rust/actions/runs/712016668 is working, other than things that are broken for other reasons. Without this fix I have to remove` stm32f7xx-hal` because it breaks all the other crates.

[...]

PR #115, opened 2021-04-02
changed to bxcan >=0.4,  <0.6

pdgilbert:
This PR is to relax the version requirement for `bxcan`. See discussion at https://github.com/stm32-rs/stm32f7xx-hal/issues/114#issuecomment-812658708

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ micromath = "1.0.0"
 synopsys-usb-otg = { version = "0.2.3", features = ["cortex-m"], optional = true }
 stm32-fmc = { version = "0.2.0", features = ["sdram"], optional = true }
 rand_core = "0.5.1"
-bxcan = "0.5"
+bxcan = ">=0.4,  <0.6"
 
 [dependencies.bare-metal]
 version = "0.2.4"


hannobraun:
Thanks for the pull request, @pdgilbert!

The CI build for this PR passes, with bxcan 0.5.0. In your comment you link to your own CI builds. I didn't dig into those, but presumably they build correctly with bxcan 0.4.0. This is all good.

The problem I'm seeing is that this PR adds a new configuration that we're not testing in our CI. Changes to the code in this crate that uses bxcan might expose some of the incompatibilities between bxcan 0.4 and 0.5, which could lead to build failures.

Maybe we could add a build configuration that uses bxcan 0.4.0, using something like `cargo update -p bxcan --precise 0.4.0`. Not sure if that would work though. I'll just merge this PR, as it addresses a real problem. But if anyone wants to look into the CI issue, that would be very welcome.

hannobraun:
I just looked into the [bxcan changelog](https://github.com/stm32-rs/bxcan/blob/master/CHANGELOG.md) and it seems like the only change from 0.4 to 0.5 is the defmt dependency. So I guess my concern is unfounded, and there's probably no way for us to introduce any bxcan-related breakage.

PR #124, opened 2021-06-18
Over-drive mode configuration

msmouni:
According to the STM32f7 data sheet, the over-drive mode should be configured this way ( Power controller (PWR) > Entering Over-drive mode ).


hannobraun:
Thank you, @msmouni! Have you tested this? Happy to merge, if you verified that it works on hardware.

msmouni:
I have tested this code on a STM32f767.

hannobraun:
Great, thank you @msmouni!

PR #125, opened 2021-06-18
Set PLLM, PLLN and PLLP

msmouni:
This can be useful if we decide to set the values in order to reduce the computation time due to calculate_mnpq() function. (Clocks.freeze() computation's time reduced by approximately 1s)

### DIFF
--- a/src/rcc.rs
+++ b/src/rcc.rs
@@ -638,6 +638,19 @@ impl CFGR {
             self.use_pll = false;
             return;
         }
+        
+        let p_ok = (sysclk as u64)
+            == (base_clk as u64 * self.plln as u64
+                / self.pllm as u64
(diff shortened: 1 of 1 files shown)


hannobraun:
Thank you, @msmouni! Looks like there's a problem with formatting (the rustfmt check failed; other CI checks are still in progress as I'm writing this).

I'm not familiar with that code, so I'm no I don't really feel qualified to review it. Maybe some comments explaining the `p_ok` calculation would help.

But I'm happy to merge, if this works for you and no one objects. Any thoughts, @mvertescher?

hannobraun:
Okay, this has been open long enough, and nobody objected. Merging!

msmouni:
Thank you

PR #127, opened 2021-07-07
Configure Git to ignore VS Code configuration

hannobraun:
An alternative would be to add it to the repository, but that's probably
too opinionated, given this crate's wide variety of configurations.


hannobraun:
@mvertescher Any objections to merging this?

mvertescher (approved):
LGTM!

hannobraun:
Thank you for the reviews, @mvertescher!

PR #128, opened 2021-07-07
Panic instead of entering endless loop on error

hannobraun:
It might be even better to return an explicit error, depending on how
the method is used, but a panic doesn't require changes to the method
signature and is a clear improvement.


hannobraun:
@mvertescher Any objections to merging this?

mvertescher (approved):
LGTM!

PR #132, opened 2021-07-13
API for the Analog to Digital converter

msmouni:
This is an API for the STM32f7xx ADC. It was tested for an STM32f767 and STM32F769.

### DIFF
--- a/src/adc.rs
+++ b/src/adc.rs
@@ -1,0 +1,562 @@
+//! # API for the Analog to Digital converter
+
+#![allow(dead_code)]
+
+use crate::rcc::{Clocks, APB2};
+
+use crate::gpio::Analog;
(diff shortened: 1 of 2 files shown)


src/adc.rs
```diff
@@ -0,0 +1,562 @@
...
+                fn channel() -> u8 { $chan }
+            }
+        )+
+    };
+}
+
+// See "Datasheet - production data"
+// Pinouts and pin description (page 66..)
+adc_pins!(ADC1,
+    gpioa::PA0<Analog> => 0_u8,
```
hannobraun, line 107: Since the type is known within the macro-generated code, it's probably unnecessary to have the `_u8` on all these numbers. Doesn't really matter though.

hannobraun (approved):
Thank you @msmouni, looks good! A bit more macro-heavy than I'd like, but it'll do fine.

PR #133, opened 2021-07-15
Fix CI build warnings

hannobraun:


src/i2c.rs
```diff
@@ -277,6 +275,12 @@ fn calculate_timing(
     an_filter: bool,
     dnf: u8,
 ) -> I2cTiming {
+    // We get an "unused" warning here when running `cargo test`. I haven't
+    // looked into it, but I suspect that is only used when `std` isn't
+    // available, which would explain that behavior.
```
rfuest, line 280: This problem is mentioned in the `micromath` docs: https://docs.rs/micromath/1.1.0/micromath/#unused-import-warnings-when-linking-std

hannobraun: Thanks, @rfuest! I updated the comment to address that.

hannobraun:
Rebased on latest master and pushed a new commit that updates a comment.

mvertescher (approved):
Looks good!

PR #134, opened 2021-07-15
Quadrature Encoder Interface API

msmouni:
This is a QEI API. It has been tested for STM32F767.


hannobraun (approved):
Thank you, @msmouni, looks great!

Issue #136, opened 2021-07-16
Clocks can no longer be specified in MHz


rfuest:
Before the migration to `embedded-time` it was possible to write `HSEClock::new(216.mhz(), HSEClockMode::Oscillator)`. This isn't possible anymore, because `Into<Hertz>` isn't implemented for `embedded-time`s `Megahertz`.

The way the conversion is implemented in `embedded-time` makes more sense, because it prevents possible integer overflows. But the use of `Into<Hertz>` in e.g. `HSEClock::new` suggests that values with any frequency unit could be used. I didn't expect this behavior while porting to the new version and had to look into why this didn't work anymore.

Would it be better to change the API to take `Hertz` values instead of `impl Into<Hertz>` values? This way the types should make the user aware that they need to convert the value into hertz first if they want to use another unit (e.g. `216.MHz().try_into().unwrap()`).

Another option would be to change the type bounds from `Into<Hertz>` to `TryInto<Hertz>`, but this would mean that all functions that take a frequency would need to return a `Result`. I don't think this would be a good idea.

hannobraun:
I agree with your conclusion that taking a `Hertz` value directly would be better and less confusing. The conversion to `embedded-time` was a rather minimally invasive procedure, and I didn't think a lot about how it affected usability.

I'd definitely merge a PR making the change from `Into<Hertz>` to `Hertz`!

hellow554:
While #137 was merged and resolves the issue with `rcc.rs`, other places are left:

```
rg 'Into<Hertz>' -n

src/sdmmc.rs:502:                pub fn init_card(&mut self, freq: impl Into<Hertz>) -> Result<(), Error> {
src/sai/pdm.rs:183:        T: Into<Hertz>;
src/sai/pdm.rs:201:                    T: Into<Hertz>,
src/sai/i2s.rs:357:        T: Into<Hertz>;
src/sai/i2s.rs:369:        T: Into<Hertz>;
src/sai/i2s.rs:388:                    T: Into<Hertz>,
src/sai/i2s.rs:411:                    T: Into<Hertz>,
src/i2c.rs:110:        F: Into<Hertz>;
src/i2c.rs:119:        F: Into<Hertz>;
src/i2c.rs:297:                    F: Into<Hertz>,
src/i2c.rs:548:                    F: Into<Hertz>
src/i2c.rs:570:                    F: Into<Hertz>
src/pwm.rs:979:        T: Into<Hertz>;
[...]

PR #137, opened 2021-07-17
Change Into<Hertz> to TryInto<Hertz> in rcc.rs

rfuest:
In contrast to my suggestion in #136 I decided to change `Into<Hertz>` to `TryInto<Hertz>` in `rcc.rs`. All methods that take a frequency argument already contained assertions that would cause a panic if the value is out of range. The same assertions can also catch if `TryInto<Hertz>` fails.

I've also added some docs to make the user aware that these methods might panic and under which conditions.

### DIFF
--- a/examples/blinky_delay.rs
+++ b/examples/blinky_delay.rs
@@ -22,7 +22,7 @@ fn main() -> ! {
     let rcc = p.RCC.constrain();
 
     // Configure clock and freeze it
-    let clocks = rcc.cfgr.sysclk(216_000_000.Hz()).freeze();
+    let clocks = rcc.cfgr.sysclk(216.MHz()).freeze();
 
     // Get delay provider
     let mut delay = Delay::new(cp.SYST, clocks);
--- a/src/ltdc.rs
+++ b/src/ltdc.rs
@@ -85,7 +85,7 @@ impl<T: 'static + SupportedWord> DisplayController<T> {
         // Get base clock and PLLM divisor
         let base_clk: u32;
         match &hse {
-            Some(hse) => base_clk = hse.freq,
+            Some(hse) => base_clk = hse.freq.0,
             // If no HSE is provided, we use the HSI clock at 16 MHz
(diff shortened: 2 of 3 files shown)


hannobraun (approved):
Thank you, @rfuest. I agree with your reasoning regarding `TryInto`.

I see you're a following the school of "first lines of documentation need to end with a period, even though that doesn't make sense because they're titles and not sentences". I won't hold that against you :grin:

rfuest:
> I see you're a following the school of "first lines of documentation need to end with a period, even though that doesn't make sense because they're titles and not sentences". I won't hold that against you grin

:grin: I try to follow the precedence set by the `std` library docs, the API guidelines and RFC 1574. It might not make sense, but at least it's consistent :wink:

hannobraun:
Ah, I didn't know this was the officially recommended style. Oh well, looks like I'm on the wrong side of history, yet again :grin:

hellow554:
I really would like to see a point release, because without this all builds fail that try to use a function that has a `{Try,}Into<Hertz>` constraint:

```
error[E0277]: the trait bound `Hertz: From<Megahertz>` is not satisfied
  --> src/main.rs:14:10
   |
14 |         .sysclk(216_u32.MHz())
   |          ^^^^^^ the trait `From<Megahertz>` is not implemented for `Hertz`
   |
   = help: the following implementations were found:
             <Hertz<u64> as From<Centihertz>>
             <Hertz<u64> as From<Decihertz>>
             <Hertz<u64> as From<Hertz>>
             <Hertz<u64> as From<Kibihertz>>
           and 5 others
   = note: required because of the requirements on the impl of `Into<Hertz>` for `Megahertz`
```

hannobraun:
Thanks for the note, @hellow554. I'm busy right now, but I've added releasing a new version to my task list. I hope I'll be able to get to it within the next 1-2 weeks.

hannobraun:
@hellow554: Version 0.5 has been published.

PR #142, opened 2021-09-13
Make `qspi.rs` DMA APIs Non-Blocking

bbrown1867:
The `qspi.rs` driver was initially written somewhat hastily and the API was a bit different than other bus drivers like `serial.rs` and `spi.rs`. This PR modifies the API to provide some consistency with the other drivers - the biggest change is making the API non-blocking: This allows the caller to decide when to `wait` for the DMA transfer to finish. 

Also APIs are renamed to `write_all`, `read_all` for DMA and simply `write` and `read` for polling variants.

The example is modified to accommodate this.

### DIFF
--- a/examples/stm32f7disco-qspi-flash/main.rs
+++ b/examples/stm32f7disco-qspi-flash/main.rs
@@ -9,6 +9,7 @@
 
 extern crate panic_semihosting;
 
+use core::pin::Pin;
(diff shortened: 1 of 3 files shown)


hannobraun (approved):
Thank you, @bbrown1867! Looks good to me.

PR #145, opened 2021-10-20
port gpio from f4-hal

burrbull:


burrbull:
> Can you confirm that this works?

No, I can't. This works on F4 and F1 good enough, but I don't have any F7 device to test.

Issue #150, opened 2021-10-27
Release 0.6.0


systec-ms:
I would like to kindly ask to publish a new version.

Mainly because of the following two pull requests, which were merged.
https://github.com/stm32-rs/stm32f7xx-hal/pull/145
https://github.com/stm32-rs/stm32f7xx-hal/pull/149

By the way, thanks @hannobraun.

PR #152, opened 2021-10-31
rcc enable

burrbull:


hannobraun:
Thank you for the pull request, @burrbull. Since this is a bit longer, I don't have time to review it immediately. I will try to take a look soon. Maybe after I release the next version (#150), I'm not sure yet.

Looks like there's a conflict now, probably due to merging #151. Can you take a look?

burrbull:
Rebased

burrbull:
I would say you need to decide what API you want.
Now all peripheral init functions are different.
Some take `&mut APBx`, other `&mut Rcc`. Some take nothing and make peripheral enable in unsafe, other in `interrupt::free` mode.

hannobraun:
@burrbull 

> Rebased

Thank you!

> I would say you need to decide what API you want.

Not sure, to be honest. It's been a while since I actually used this HAL, or worked on the design of a STM32 HAL, or really thought about the design of STM32 HALs.

It would be better if it was consistent, of course, but I don't know which of the options is best. Suggestions welcome!

hannobraun (approved):
Thank you, @burrbull, this looks great! I think that new infrastructure in RCC is a nice improvement.

You mentioned last time that you don't have F7 hardware. I assume this is still true, and this change hasn't been tested. Since this affects pretty much everything, I'd like to see some testing done before merging. As usual, it might take a while before I can get to it myself.

If anyone is reading this and wants to help out, please run some examples or your own code, and let us know how it works for you!

hannobraun:
As for the inconsistencies, I've opened https://github.com/stm32-rs/stm32f7xx-hal/issues/154.

systec-ms:
LGTM, tested with a STM32F767.

src/rcc/enable.rs
```diff
@@ -0,0 +1,219 @@
...
+    SPI5 => (APB2, spi5en, spi5lpen, spi5rst), // 20
+    SAI1 => (APB2, sai1en, sai1lpen, sai1rst), // 22
+    SAI2 => (APB2, sai2en, sai2lpen, sai2rst), // 23
+}
+
+#[cfg(any(feature = "svd-f730", feature = "svd-f7x2", feature = "svd-f7x3",))]
+bus! {
+    AES => (AHB2, aesen, aeslpen, aesrst), // 4
+
+    SDMMC2 => (APB2, sdmmc2en, sdmmc2lpen, sdmmc2rst), // 7
```
systec-ms, line 170: The `SDMMC2` is not only available in `svdf7x2` and `svd-f7x3`:

``` bash
stm32f7xx/svd$ rg SDMMC2 -l --sort-files
STM32F7x2.svd
STM32F7x3.svd
STM32F7x7.svd
STM32F7x9.svd
```

burrbull: `f765`, `f7x7` and `f7x9` doesn't have `sdmmc2en` and `sdmmc2rst` fields in `RCC`.
This should be fixed in `stm32-rs` first

hannobraun:
Thanks for testing, @systec-ms, and thanks again for the pull request, @burrbull!

PR #155, opened 2021-11-04
Add I2C scanner example

systec-ms:
Inspired from [stm32f3xx-hal](https://github.com/stm32-rs/stm32f3xx-hal/blob/master/examples/i2c_scanner.rs.)

Tested with a STM32F767.

### DIFF
--- a/examples/i2c_scanner.rs
+++ b/examples/i2c_scanner.rs
@@ -1,0 +1,61 @@
+//! Example of using I2C.
+//! Scans available I2C devices on bus and print the result.
+
+#![no_std]
+#![no_main]
+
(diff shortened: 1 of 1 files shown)


hannobraun (approved):
Thank you, @systec-ms!

Looks good. The only note I have is that I wouldn't consider semihosting a good option anymore. It would probably be best to replace it with something based on RTT, possibly defmt, and replace the whole OpenOCD configuration in this repository with cargo-embed or probe-run.

But yeah, that's out of scope for this pull request, I'd say :smile:

hannobraun:
By the way, your link doesn't work (errant `.` at the end).

PR #156, opened 2021-11-04
Fix RAM address and add ITCM and DTCM sections

### LINKED ISSUE #123: Error in memory.x ?
On line 6 of memory_2048_368.x there is 
"RAM : ORIGIN = 0x20000000, LENGTH = 368K + 16K"
according to datasheet it would be 
"RAM : ORIGIN = 0x20000000, LENGTH = 128K + 368K + 16K"
or
"RAM : ORIGIN = 0x20020000, LENGTH = 368K + 16K"

systec-ms:
Previously DTCM region was simply added to RAM region, this can cause problems with DMA as DMA does not have access to DTCM region.
See page 71 and 76 in reference manual RM0410, Rev 4.

Also https://github.com/stm32-rs/stm32f7xx-hal/issues/123

Signed-off-by: Moritz Scheuren <moritz.scheuren@systec-electronic.com>

edit by @hannobraun: Make sure we "close #123" automatically once this merges.

### DIFF
--- a/memory_2048_368.x
+++ b/memory_2048_368.x
@@ -3,8 +3,31 @@ MEMORY
 {
   /* NOTE K = KiBi = 1024 bytes */
   FLASH : ORIGIN = 0x08000000, LENGTH = 2M
-  RAM : ORIGIN = 0x20000000, LENGTH = 368K + 16K
+  RAM : ORIGIN = 0x20020000, LENGTH = 368K + 16K
+  ITCM : ORIGIN = 0x00000000, LENGTH = 16K /* Instruction Tighly Coupled Memory */
+  DTCM : ORIGIN = 0x20000000, LENGTH = 128K /* Data Tighly Coupled Memory */
 }
+
+SECTIONS
+{
+    .itcm : ALIGN(4)
+    {
+        *(.itcm .itcm.*);
+        . = ALIGN(4);
+    } > ITCM
+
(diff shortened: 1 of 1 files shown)


hannobraun (approved):
Thank you, @systec-ms, looking good!

Are you happy with your testing of this change, or do you think this could use additional testing? Happy to merge this, if you are, I just won't have time to do much testing myself any time soon.

systec-ms:
Thank you and good question.
The ITCM and DTCM should not be used by default, see ([link.x.in](https://github.com/rust-embedded/cortex-m-rt/blob/master/link.x.in)), so the whole thing should not cause problems for now.
Until someone explicitly uses ITCM or DTCM with something like `#[link_section = ".dtcm.BUFFER"]`.

Regarding the two, I tested the following:
- [x] DMA access of a `BUFFER` in this region leads to a crash (as expected)
- [x] `cargo size --release --example test -- -A`, looks like `BUFFER` was loaded into the appropriate region
- [x] `cargo objdump --release --example test -- -D  | less`, looks like `BUFFER` was loaded into the appropriate region

I would leave it up to you to decide if all of this is sufficient or if you have any more ideas on what to test.

hannobraun:
Thanks, @systec-ms, that sounds great. Especially since it's an opt-in feature.

Merging.

PR #158, opened 2021-11-18
Bump bxcan version from ">=0.4,  <0.6" to "0.6"

systec-ms:


hannobraun (approved):
Thank you, @systec-ms!

PR #159, opened 2021-11-26
RTC: Add hse and lsi as clock source

systec-ms:
Motivation: I need an RTC and I don't have a `LSE`, so I want to use the RTC with the `HSE`.

Tested:
- [x] HSE
- [x] LSI
- [ ] LSE (I don't have one)

Clocks structure for the `Rtc::new` function would not be absolutely necessary at the moment, but allows to implement a functionality later, which calculates the prescaler from a given frequency (not part of this PR).

It would also be debatable whether to wait for `LSI` and `LSE` inside the freeze function, or to do so where they are needed.

I have previously looked at the `H7-HAL` and was slightly inspired by it.


hannobraun (approved):
Thank you, @systec-ms, looks reasonable!

PR #160, opened 2021-12-08
RTC timestamp, was not read atomic, which can lead to wrong timestamps.

systec-ms:


hannobraun (approved):
Thank you, @systec-ms! Looks great, as always.

adamgreig:
I saw this PR just after writing a bunch of RTC handling code for another project so maybe it will be useful:

Instead of returning an error, maybe it would be nicer to perform a second read and so always return consistent data?

The RM says (in RM0385  29.3.8 Reading the calendar) that if the APB clock is >= 7x the RTC clock (which is very likely since RTC clock is usually 32kHz) then it's OK to just read TR and then DR: the hardware locks DR after you read TR to ensure you get consistent data. In fact, it's important to always read TR and then DR, otherwise the later read of TR will lock an old value of DR. But, even if the APB clock is extremely slow, you can always read TR, then DR, then TR, then DR, and use the first values if the two TRs match, or the second values if the TRs don't match (as also described in the RM). Seems friendlier than returning an error to the user who has no choice but to now re-read in a loop until it doesn't error.

It looks like the driver should also wait for RSF to be set before taking values in case the user re-reads very soon afterwards too, and clear it after each read. Maybe it would be best to have a single `get_datetime()` method that does the correct wait-for-RSF, read TR, read DR, clear RSF sequence, and parses the fields, and then convenience methods for reading just seconds/minutes/etc could call that and return the relevant bit. Or just always give the user a full consistent datetime to make sure they don't try and read each part separately and end up with inconsistent data that way anyway...

systec-ms:
> Instead of returning an error, maybe it would be nicer to perform a second read and so always return consistent data?

Probably yes, I did it that way since the error type was already there.

> It looks like the driver should also wait for RSF to be set before taking values in case the user re-reads very soon afterwards too, and clear it after each read.

You are right.

> Maybe it would be best to have a single get_datetime() method that does the correct wait-for-RSF, read TR, read DR, clear RSF sequence, and parses the fields, and then convenience methods for reading just seconds/minutes/etc could call that and return the relevant bit. Or just always give the user a full consistent datetime to make sure they don't try and read each part separately and end up with inconsistent data that way anyway...

I agree, but I am unsure which one is better. Returning always a full consistent datetime would conflict a bit with the rtcc trait. Nevertheless, I currently think this would be the best option, and you rebuild the whole thing on the [chrono crate](https://crates.io/crates/chrono).


Also, that every get function sets the calendar clock to 24 hr format with `self.set_24h_fmt();` is not superb.

adamgreig:
It might also be worth checking the (revamped) [time crate](https://crates.io/crates/time) which works fine on no_std now and seems more currently maintained, though I'm not sure what the general backstory is between them.

hannobraun:
Thank you for your insight, @adamgreig. I like the idea of only returning a full date/time to ensure consistency. I don't know how whether it's worth diverging from the `rtcc` traits for this, but maybe an argument could be made to change the `rtcc` traits.

In any case, I don't have time to implement any of the suggestions (as usual), but I'm happy to review and merge pull requests, in case anyone wants to make further improvements here.

systec-ms:
See https://github.com/stm32-rs/stm32f7xx-hal/pull/162.

PR #164, opened 2022-01-05
rcc: Add pllsai to CFGR struct and pllsaip as PLL48CLK source

systec-ms:


src/rcc.rs
```diff
@@ -967,9 +1051,35 @@ impl CFGR {
...
+            };
+            let pllsaip_freq = pllsain_freq
+                / match self.pllsaip {
+                    PLLSAIP::Div2 => 2,
+                    PLLSAIP::Div4 => 4,
+                    PLLSAIP::Div6 => 6,
+                    PLLSAIP::Div8 => 8,
+                };
+            // let pllsaiq_freq = pllsain_freq / self.pllsaiq as u64;
+            assert!((192_000_000..=432_000_000).contains(&pllsain_freq));
```
systec-ms, line 1067: I'm not sure here, the reference manual (RM0410 Rev 4, Page 212), says the following `Caution: The software has to set these bits correctly to ensure that the VCO output frequency
is between 100 and 432 MHz.`, but STM32CubeMX states 192 MHz as the minimum.

src/rcc.rs
```diff
@@ -457,7 +483,41 @@ impl CFGR {
...
+    pub fn use_pllsai(mut self) -> Self {
+        self.use_pllsai = true;
+        self
+    }
+
+    /// Sets the PLLSAIN multiplication factor for PLLSAI.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the multiplication factor isn't between 50 and 432.
```
hannobraun, line 496: Possible slight clarification:
```suggestion
    /// Panics if the multiplication factor isn't between 50 and 432 (inclusive).
```

src/rcc.rs
```diff
@@ -457,7 +483,41 @@ impl CFGR {
...
+    pub fn pllsaip(mut self, pllsaip: PLLSAIP) -> Self {
+        self.pllsaip = pllsaip;
+        self
+    }
+
+    /// Sets the PLLSAIQ division factor for PLLSAIS.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the division factor isn't between 2 and 15.
```
hannobraun, line 513: Same as above:
```suggestion
    /// Panics if the division factor isn't between 2 and 15 (inclusive).
```

src/rcc.rs
hannobraun: Could you please add that exact comment as a comment to the code?

hannobraun (changes requested):
Thank you, @systec-ms!

I did a quick review and found nothing unreasonable. I don't have time to check everything against the Reference Manual, but assuming you tested this, I'm fine with merging.

I've left some suggestions on making some documentation (that is already pretty clear) a tiny bit more precise. No strong opinion here, take it or leave it.

The only change I'd like to see before merging, is to add that comment you added here on GitHub to the source code instead. I think there it'll be more useful for making sense of the code later.

systec-ms:
Thanks, I tested the clock output of the `PLL48CLK` with `pllsaip` as the clock source.
Btw, should I squash?

hannobraun (approved):
Thanks again, @systec-ms, looks great now!

> Btw, should I squash?

I was about to say, "nah, it's fine", but then you got me thinking that these three commits actually belong together :grin: 

Luckily, there's the "squash and merge" option, so no reason for another round of change and review.

(As a general rule, I prefer new commits being added while a pull request is open, as that makes reviewing new changes much easier. And I'd rather have too many commits than too few. Too many commits can get unwieldy, but few and huge commits are much worse from a review perspective.)

PR #167, opened 2022-01-28
Add initial PWM support

maximeborges:
The main purpose of this PR is to get some initial support for PWM.
It also integrates some features for timers from the `stm32f4xx-hal` repo.
This will break existing implementation, but get us an much cleaner interface for instantiating and using timers.

PR #168, opened 2022-01-28
Add support for flash.rs on bigger chips

maximeborges:
The current code to erase flash is only using the `mer` field.
On bigger chips, there is two memory banks, so the `mer` field is split in two for the two memory banks, `mer1` and `mer2`.

This PR add checks to configure either `mer` or `mer[12]` for erase functions.

### DIFF
--- a/src/flash.rs
+++ b/src/flash.rs
@@ -180,9 +180,7 @@ impl<'a> EraseSequence<'a> {
                 feature = "stm32f778",
                 feature = "stm32f779",
             ))]
-            w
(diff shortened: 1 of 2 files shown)


hannobraun:
See https://github.com/stm32-rs/stm32f7xx-hal/pull/167#issuecomment-1024871906.

maximeborges:
I confirm that this example works on my Nucleo-F767ZI

hannobraun (approved):
CI is green now. Thank you, @maximeborges!

Issue #163, opened 2021-12-10
Timer example is incorrect


systec-ms:
If the interrupt is not cleared, it will immediately retriggered after the ISR has finished.
``` rust
    timer.clear_interrupt(Event::TimeOut);
```

hannobraun:
Good catch, thank you!

systec-ms:
@hannobraun could you add `good first issue` as a label, I can't. Thanks.

hannobraun:
Good call!

darryln:
@systec-ms  is there a repro case, or was this found by inspection?

systec-ms:
> @systec-ms is there a repro case, or was this found by inspection?

Inspection.

eldruin:
Fixed in #169. Closing.

PR #169, opened 2022-01-28
example/timer: Clear the interrupt flag

maximeborges:
Closing https://github.com/stm32-rs/stm32f7xx-hal/issues/163

### DIFF
--- a/examples/timer.rs
+++ b/examples/timer.rs
@@ -6,8 +6,8 @@
 
 extern crate panic_semihosting;
 
+use core::cell::RefCell;
 use core::fmt::Write;
-use core::cell::{RefCell};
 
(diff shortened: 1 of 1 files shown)


hannobraun:
See https://github.com/stm32-rs/stm32f7xx-hal/pull/167#issuecomment-1024871906.

hannobraun (changes requested):
Thank you, @maximeborges, looks good! (Although in anything that goes beyond a simple example, I'd prefer RTIC for this kind of thing.)

CI fails due to formatting. A `cargo fmt` should do it. Happy to merge, once CI is green.

hannobraun (approved):
CI is green. Thanks, @maximeborges!

PR #171, opened 2022-02-11
allow to configure:

msmouni:
+UsartInversion
+WordLength
+ParityControl

PR #172, opened 2022-02-11
allow struct Serial to do DMA Transfers, 

msmouni:
allow struct Serial to do DMA Transfers without calling split(), in order to allow releasing the ressources after using DMA transfers


hannobraun:
See https://github.com/stm32-rs/stm32f7xx-hal/pull/171#issuecomment-1036339267.

PR #176, opened 2022-03-02
update gpio

burrbull:
cc @eldruin


eldruin (approved):
Wow, what a huge amount of work!
Thank you and take care @burrbull!

PR #179, opened 2022-03-10
Port timers from f4xx-hal

burrbull:


eldruin (approved):
Great whole bunch of work. Thank you!

PR #180, opened 2022-04-14
port watchdog

ahdinosaur:
hi stm32f7 folks! :smiley_cat: 

here's a pull request to port a watchdog from [stm32f4xx-hal](https://github.com/stm32-rs/stm32f4xx-hal/blob/70894b4/src/watchdog.rs) and [stm32f1xx-hal](https://github.com/stm32-rs/stm32f1xx-hal/blob/f9b24f4/src/watchdog.rs).

cheers :seedling:


eldruin (approved):
Looks good to me, thank you!

PR #182, opened 2022-05-23
Serial data bits and default config for serial examples

maximeborges:
Following #181 to add control over parity on UARTs, adding control of data bits.
Also using default config in examples to avoid having to change them every time there is a modification in the config structure.


maximeborges:
@ahdinosaur could you check if this PR doesn't break anything on your side ?

ahdinosaur:
hi @maximeborges, yep looks good here! :smiley_cat:

PR #189, opened 2022-06-05
Add config for VSCode rust-analyzer

maximeborges:
`rust-analyzer` on VSCode requires selecting a MCU feature to run.

This PR adds a default config with some info in the README to tell user to change it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,3 @@
 **/*.orig
 **/*.rs.bk
 Cargo.lock
-
-# VS Code
-/.vscode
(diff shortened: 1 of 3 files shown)


eldruin (commented):
Good idea!
However, I would rather keep the .vscode folder in the .gitignore so that when people change the MCU and features, it does not annoy everybody with the pending changes.
Other than that, great!

maximeborges:
It looks like having the file in .gitignore doesn't change anything if it's tracked by git anyway.
I propose to remove it so the .gitignore file stays as clean as possible related to what files are actually tracked.

eldruin (approved):
Ah, my bad. Thanks for checking on it.

PR #190, opened 2022-06-05
README: rewrite based on stm32f1xx-hal

maximeborges:
Since the current README is a bit empty and not really up to date anymore concerning the main sources of inspiration, this PR refresh that a bit, based on the [stm32f1xx-hal](https://github.com/stm32-rs/stm32f1xx-hal).
Ideally it'd be nice to have some examples on how to use this lib like for the above-mentionned HAL, but setting up `probe-run` as default runner first might get us a more accessible and up-to-date environment to write that on top off.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,19 +1,59 @@
 # `stm32f7xx-hal`
 
+> [HAL] for the STM32F7 family of microcontrollers
+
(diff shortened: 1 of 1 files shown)


eldruin (approved):
Great, thank you!

PR #192, opened 2022-07-24
Small documentation updates

sveinrou:
This fixes a broken link in the rtc-docstring, as well as removing the reference to F303 section 27

Also, while I was at it, I added a short docstring to the modules that didn't have it, so the module index on the docs front page looks a little nicer


eldruin (approved):
Looks good to me, thank you!

PR #194, opened 2022-12-07
Bump bxcan

andyblarblar:
Bxcan is on v0.6 here, v0.7 adds embedded-hal traits, which cannot be used with this crate since it is on v0.6. 

This PR simply updates bxcan.


eldruin:
Great, thanks!
Could you fix the example?

andyblarblar:
> Great, thanks! Could you fix the example? The API seems to have changed slightly.

Yep, forgot about that. It should be all fine now.

eldruin (approved):
Great, thank you!

PR #195, opened 2023-05-18
examples: blink-timer-irq: fix clock panic

hexagonal-sun:
The example sets the frequency of the pclk1 to an out-of-bounds frequency which causes the example to panic in all instancies.

Set to a frequency that is in-range.

Tested on: STM32F746G-DISCO


eldruin (approved):
Nice find and thank you for the fix!

PR #197, opened 2023-06-08
Enabled interruptable gpio on alternate function pins

Cryowatt:
I noticed while trying to use the `ExtiPin` trait on an `Alternate` function pin that the `ExtiPin` trait wasn't enabled. Based on [examples from other STM32 hals](https://github.com/stm32-rs/stm32f4xx-hal/blob/43a97678f8772af1712fec3af7b3391cf02f6ba6/src/gpio.rs#L173) I impl'd the trait. It appears to work correctly on my STM32F723.

closes https://github.com/stm32-rs/stm32f7xx-hal/issues/198


eldruin (approved):
Great, thank you!

PR #199, opened 2023-08-09
Add a "low pin count" variant of f730

BryanKadzban:
The f730 devices don't all have a USB-high-speed PHY, only the 144-pin and larger packages have it.  Don't force the PHY for all 730 devices.


eldruin (commented):
Thank you! Seems fine to me.
Could you add an entry about this to the changelog?

BryanKadzban:
Done, and squashed into the same commit.

eldruin (approved):
Thank you!

PR #203, opened 2023-10-02
serial: set active state level for RX and TX pins

maximeborges:
Fresh take on https://github.com/stm32-rs/stm32f7xx-hal/pull/171, rebased on latest changes


eldruin (commented):
Thanks! could you have a look at the CI errors?

PR #204, opened 2023-10-04
Fix build.rs / lib.rs after b497ddec

BryanKadzban:
I never actually tried to build.  :-/  Add support for the -lpc variant to build.rs and the module definitions in lib.rs.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -9,6 +9,7 @@ compile_error!(
         stm32f722
         stm32f723
         stm32f730
+        stm32f730-lpc
         stm32f732
         stm32f733
         stm32f745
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -34,6 +35,9 @@ pub use stm32f7::stm32f7x3 as pac;
 #[cfg(feature = "stm32f730")]
 pub use stm32f7::stm32f730 as pac;
 
+#[cfg(feature = "stm32f730-lpc")]
+pub use stm32f7::stm32f730 as pac;
```
maximeborges, line 39:
```suggestion
#[cfg(any(feature = "stm32f730", feature = "stm32f730-lpc"))]
pub use stm32f7::stm32f730 as pac;
```

BryanKadzban: Makes sense -- done.

maximeborges (approved):

PR #206, opened 2023-10-27
Add UART8 and rearrange definitions of pins

rtvd:
Pin definitions for UART8 were missing.

Also, as pins were grouped by type  rather than by UART it was not easy to match their definitions to the datasheet.
The change was done in two steps: first I have re-arranged the definitions of pins and grouped them by UART (on the high level) and then by their location in the datasheet, then I have added the pins of UART8.


rtvd:
@maximeborges please could you take a look? I hope it is not too awful.

maximeborges:
I would prefer to group the pins by UART and then by type instead of having them by UART and then by "pair", since there is no concept of RX/TX pins pair AFAIK, i.e.:

```rust
// USART1
impl PinTx<USART1> for gpio::PA9<Alternate<7>> {}
impl PinTx<USART1> for gpio::PB6<Alternate<7>> {}

impl PinRx<USART1> for gpio::PA10<Alternate<7>> {}
impl PinRx<USART1> for gpio::PB7<Alternate<7>> {}
```

instead of:

```rust
// USART1
impl PinTx<USART1> for gpio::PA9<Alternate<7>> {}
impl PinRx<USART1> for gpio::PA10<Alternate<7>> {}

impl PinTx<USART1> for gpio::PB6<Alternate<7>> {}
impl PinRx<USART1> for gpio::PB7<Alternate<7>> {}
```

rtvd:
OK.  Updated.

maximeborges (approved):

PR #210, opened 2024-02-08
Fix mass-erase trigger on "small" devices.

BryanKadzban:
The list of specifically enumerated "large" devices has the correct set of register operations to start a mass erase, setting the mer1 and mer2 bits.  But the "small-devices" alternative does not; it clears the mer bit instead of setting it.

Looking at history, this was changed from set to clear in the PR that added support for "large" devices.  It appears to have been accidental.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added Interruptable trait to Alternate mode pins
 - Added a "low pin count" variant of the f730 chip to the crate features: packages <144 pins don't include a high speed USB PHY
 - Added SPI2_SCK pin for stm32f769i-discovery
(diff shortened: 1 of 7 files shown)


BryanKadzban:
The CI tests are failing due to the nightly compiler change to reject all references to mutable statics, not due to this change.

eldruin (commented):
I see, thanks!
Could you add an entry to the changelog about this?
Could you also add the couple of `addr_of_mut!` calls so that the nightly compiler is happy?

BryanKadzban:
> Could you add an entry to the changelog about this?

Done.

> Could you also add the couple of `addr_of_mut!` calls so that the nightly compiler is happy? (this does not need a changelog entry)

Also done.

Thanks!

eldruin (approved):
Great, thank you!

PR #212, opened 2024-02-11
Use fixed-point u64 math to divide the PLL values

### LINKED ISSUE #211: Clock setup code pulls in almost 1K of u64-math builtins
I'm trying to write a bootloader, meaning the output code needs to be as small as possible.  In looking at the big symbols being linked into the output binary, I see that compiler_builtins::int::specialized_div_rem::u64_div_rem::[hash] is almost 800 bytes, and it's only called from a couple builtins (__aeabi_uldivmod calls __udivmoddi4 which calls it).  Those builtins are 24 and 22 bytes, for a total of 842.  This is uncomfortably close to 1K.  There is only one bigger chunk, in all the core::fmt support code, totaling over 4K -- removing that (hopefully defmt is better) is a bigger change.

The only thing in the bootloader that calls __aeabi_uldivmod is the clock setup code.  It does it in three places; two which refer to this chunk of source from "cargo objdump --release -- -d --demangle --source":

sysclk = base_clk as u64 * self.plln as u64 / self.pllm as u64

and one that refers to this one:

let f_vco_clock = (f_pll_clock_input as u64 * n as u64 / m as u64) as u32;

Perhaps these divide-by-m options don't need to be u64?  Or perhaps it'd be possible to open-code the 64-bit division, or *perhaps* do them with a shift instead of a divide.  (I doubt that, since I think m can be values other than powers-of-2, but perhaps.)  Or maybe represent the frequencies in kHz instead of Hz, if we're only a couple bits over the 32-bit u32 range.  Or maybe do the division first (still as a u32), to avoid overflow.

I'll be looking into this, but it'd be good to know if others are thinking about it already as well.

BryanKadzban:
This avoids the need for almost 1K of code to implement division for u64 values (which the Cortex-M7 can't do in hardware).  It can multiply u64 values in hardware, so instead of dividing by (integer) x, we find 1/x in fixed-point, multiply, and shift the fixed-point offset back out. It can also divide u32 values, and finding 1/x can be done in a u32 since our shift is small enough.

Fixes #211

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added a "low pin count" variant of the f730 chip to the crate features: packages <144 pins don't include a high speed USB PHY
 - Added SPI2_SCK pin for stm32f769i-discovery
 - Fix mass-erase triggering in `flash` on smaller chips
+- Remove the need for software u64 division in the clock setup code, shrinking code (#211)
 
 ## [v0.7.0] - 2022-06-05
 
--- a/src/rcc.rs
+++ b/src/rcc.rs
@@ -558,6 +558,33 @@ impl CFGR {
         self
     }
 
+    // We want to avoid dividing u64 values, because the Cortex-M7 CPU doesn't
+    // have hardware instructions for that, and the software divide that LLVM
+    // gives us is a relatively large amount of code.
+    //
+    // To do this, we operate in a fixed-point domain, and do a multiply by 1/x
+    // instead of dividing by x.  We can calculate those 1/x values in a u32, if
+    // the fixed-point decimal place is chosen to be close enough to the LSB.
+    //
+    // But we also need to be able to represent the largest numerator, so we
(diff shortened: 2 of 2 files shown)


BryanKadzban:
Hmm, I seem to be failing a bunch of rcc tests.  Let's fix that up...

BryanKadzban:
For people not looking at #211, the issue with the tests was that we lost too much precision using only 26 bits for the fractional part of the fixed-point math.  It needed at least 30 bits.  But the numerator in the division operation needed up to 38 bits, so it would overflow a u64.  Fortunately, we could shift the frequency right by 4 bits at the start to make everything fit, as long as the input frequencies are a multiple of 250kHz (they all are in the tests, at least, and I think they are likely to be in people's designs as well).

So now we shift the frequency right by 4 bits and use 30 bits for the fractional part.  Tests are all accurate enough to pass.

eldruin (commented):
This looks good to me in principle, but I do not know exactly how it works. If nobody raises any issues, I would merge this.

eldruin (approved):
Alright, let's merge this. Thank you for your work!

PR #216, opened 2024-09-11
Fix the addresses of `signature` module structs for some devices

BryanKadzban:
For some reason ST puts these registers at different locations on different devices.

ST publishes three different reference manuals: RM0385, RM0410, and RM0431.  The first two document the UID and FlashSize registers at the addresses used before this commit; the last has them at the newly added addresses.  That last reference manual is for the F7x2, F7x3, and F730 devices, so use these other addresses for those three types of device.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 <!-- next-header -->
 ## [Unreleased] - ReleaseDate
 
+### Fixed
+
+- Use the right addresses for `signature` module structs on f7x2, f7x3, and f730 devices
+
 ## [0.8.0] - 2024-08-02
 
 ### Changed
--- a/src/signature.rs
+++ b/src/signature.rs
@@ -30,6 +30,9 @@ pub struct Uid {
     y: u16,
     waf_lot: [u8; 8],
 }
+#[cfg(any(feature = "svd-f7x2", feature = "svd-f7x3", feature = "svd-f730"))]
+define_ptr_type!(Uid, 0x1FF0_7A10);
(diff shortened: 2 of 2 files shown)


eldruin (commented):
Thank you!
I see you have added alternative addresses for `FlashSize` and `Uid` but what about the other structs that are now conditional (`VrefCal`, `VtempCal30` and `VtempCal110`)?
Have you tested this in hardware?

BryanKadzban:
Those three structs aren't in the 0431 reference manual (none of the references to calibration are any of these registers), so yes, dropping them was intentional.

I didn't actually look to see if they're in both 0385 and 0410 though...  Hmm.  They're not in either, actually.  I wonder if that means they're documented in some other manual, and maybe they *are* present on the F722 device?  Looking at the original commit that added this signature module, it was first turned on for only the F746 chip ... aha, it's documented in the datasheet, not the reference manual (in a couple tables in the "temperature sensor" and "reference voltage" sections).  And the datasheet for the 722 chip shows the same fields are present, but also at a different address than the 746.

OK, fixing that properly will take a lot of datasheet reading (since each chip has its own datasheet; they don't share an ID like the reference manuals do), so I'll re-add it with the proper addresses for everything, and upload something new when that's done.  Might take a while.

I have tested accessing the Uid struct on a real F722 chip (that was all I needed); without this patch the code triggers a hard-fault, and with it, I get valid-looking signature data.

BryanKadzban:
OK, added the registers back for the other chips.  It turns out the address in use *does* correspond to the same groups of chips as I had already -- F7x2, F7x3, and F730 have the same address as each other, and a different address from every other F7xx chip.

So I think this will work better.

eldruin (approved):
Great, thank you!

PR #217, opened 2026-08-03
Update CAN interface remapping for RX/TX pins and add support for PD0/PD1

JoeKL:
- Add CAN1 support for `PD1` as TX and `PD0` as RX using alternate function 9.
- Document the additional CAN1 pin mapping.
- Hardware test was perfomed.
- Entry in Changelog was added


eldruin (approved):
Looks good to me, thank you!

